PR #424, opened 2014-07-14
Always keep the repository checkout up to date with GitHub

lkundrak:
Don't merge with earlier changesets -- the user might have chosen to force-push
over them. If we run git pull, it would attempt to merge the changes,
potentially resulting in a conflicted state.


henare:
This is brilliant and fixes a bug we just encountered in the wild :beetle:

Huge thank you @lkundrak! :sparkles:

henare:
At first I was nervous about using a hard reset in this PR but after a bit of experimentation with checking out origin/master in a detached head state I think this is the best solution.

PR #465, opened 2015-01-21
Give more detail for Ruby newcomers

softgrow:
Try and give enough detail so an absolute novice can get started with Ruby and morph.io


mlandauer:
Thanks @softgrow

PR #479, opened 2015-03-18
resizable console on scraper css

joonjoonjoon:
this will allow the console to be resizable


mlandauer:
Thanks so much for the PR @joonjoonjoon - really appreciate it. I'll test this on my dev machine now.

mlandauer:
I've also deployed this to production. You might need to refresh the page a bit to get the latest css. Thanks again for the fix @joonjoonjoon!

PR #539, opened 2015-04-09
be more specific about database settings

equivalentideas:
I had to ask for help to clarity what database setting were required to be changed. I'm pretty noobish with database stuff though so maybe this is unnecessary or wrong.


mlandauer:
I agree that this bit could do with some more detailed explanation. I don't think this wording change helps clarify things though

Issue #557, opened 2015-04-10
Update to bootstrap 3.3


equivalentideas:
Morph currently uses bootstrap 3.2. We should update to get the advantages of the latest version, and also change the implementation so we can change the variables.

mlandauer:
@equivalentideas is this a blocker for anything you're doing at the moment?

equivalentideas:
fixed by https://github.com/openaustralia/morph/commit/c7a52680c6fa43726f36e7cea50e70960d9c84d8

PR #560, opened 2015-04-10
Switch to bootstrap sass

mlandauer:
This changes how rails is importing bootstrap. We're now using the bootstrap sass gem and updated bootstrap to 3.3.

A quick poke around the site looks like everything is working. @equivalentideas it would be great if you could take a look and test it out. Make sure you restart the rails server when you switch to this branch.


mlandauer:
This PR fixes #557

equivalentideas:
Tried adding some scrapers and poking around. Couldn't find any problems.

Are we going to merge this into the new_landing_page branch?

mlandauer:
Thanks for testing this out @equivalentideas. I'll merge this into master and then I think we should merge master back into new_landing_page so we get the latest available there too.

Issue #588, opened 2015-04-15
make table fade out scroll affordance adapt to citizen interaction


equivalentideas:
The fade out on the side of a data table is a little tip to the user that there is more of the table off to the right.

![screen shot 2015-04-15 at 3 29 48 pm](https://cloud.githubusercontent.com/assets/1239550/7152713/4c92e1b4-e384-11e4-9204-2ab02900a055.png)

But, when they scroll over and there isn't any more to see, it is still there:

![screen shot 2015-04-15 at 3 25 44 pm](https://cloud.githubusercontent.com/assets/1239550/7152717/5eb4001c-e384-11e4-813e-7069343af368.png)

The fade should be on whichever side has more table to reveal, or both.

mlandauer:
:+1:

equivalentideas:
fixed by https://github.com/openaustralia/morph/commit/b6c322228f485c73b397dc1503040578679524f0

Issue #356, opened 2014-05-17
Make data table more flexible


mlandauer:
Looks like https://github.com/filamentgroup/tablesaw might be useful for this

equivalentideas:
There are a few different elements to this. We want to make the tables layout flexible because currently it is:

* not always obvious there is more data to view
* scrolling around tables is a very bad experience on small screens
* potentially we can save citizens from having to scroll at all if we can get more data in the initial view

equivalentideas:
For very small screens we could use [TableSaw](https://github.com/filamentgroup/tablesaw), potentially in the [Column Toggle mode](http://filamentgroup.github.io/tablesaw/demo/toggle.html), which could make the table very readable on small screens.

For mid sized screens we could adjust the max-width of columns to pack more into this view.

I think these steps plus #588 are a good start.

mlandauer:
What's the reason for not trying to use TableSaw on all screen sizes for simplicity?

equivalentideas:
Sorry that's a bit confusing, I meant that TableSaw changes the table for small screens, but it leaves widescreen tables alone—so it's only helping us there. Unless we use some of the extra features like being able to toggle columns, but I'm not so sure about that now, it could make it less clear what columns are present.

equivalentideas:
closed by https://github.com/openaustralia/morph/commit/b6c322228f485c73b397dc1503040578679524f0

PR #597, opened 2015-04-16
[WIP] Use TableSaw on scraper data tables

equivalentideas:
This is the work in progress of implementing [TableSaw](https://github.com/filamentgroup/tablesaw) on scraper data tables. There are a few bugs and todos below.

Relates to #356

TODOS:

* [x] Make the distinction between rows clearer in <dl> style
* [x] remove the righthand fadeout when in <dl> style ( relates to #588 )
* [x] **BUG** For some reason it works for some the data from some scrapers (like [/soit-sk/firebrno](https://morph.io/soit-sk/firebrno)) but not others (like [/planningalerts-scrapers/woollahra](https://morph.io/planningalerts-scrapers/woollahra) or [/planningalerts-scrapers/city_of_sydney](https://morph.io/planningalerts-scrapers/city_of_sydney)). I don't understand why this is happening yet.
* [ ] **BUG** When a scraper is run, it stops the TableSaw js working
* [ ] I think we should cut the table from 10 to 3 or 5 rows on small screens, as it can get extremely long if some of the table cells are really long. For someone on a phone, is 10 rows necessary to get an understanding of what the data is? I don't think so. We could do the hiding in css then update the row count in javascript when tablesaw is doing it's thing on small screens.
* [x] If the current table is not wider than 100%, there's no reason to set a max-table width.

Here's what it currently looks like on small screens:


equivalentideas:
@mlandauer This is ready for review, but for a bug and a question.

Firstly, what do you think about:

>  I think we should cut the table from 10 to 3 or 5 rows on small screens, as it can get extremely long if some of the table cells are really long. For someone on a phone, is 10 rows necessary to get an understanding of what the data is? I don't think so. We could do the hiding in css then update the row count in javascript when tablesaw is doing it's thing on small screens.

Secondly:

>  BUG When a scraper is run, it stops the TableSaw js working

This is because when [Sync](https://github.com/chrismccord/sync) reloads the partial, the TableSaw js and the [setScrollerFadeEffect](https://github.com/openaustralia/morph/pull/597/files#diff-c4d3b3fb784ef28b6d398f3055a56a3eR48) function, don't rerun on the new table.

There's hooks for this in the [Sync docs]() but I can't get it working, I was wondering if this was something you're already on to of or could think of another solution?

mlandauer:
Thanks @equivalentideas. I'll take a look when I get back from lunch

mlandauer:
Thanks @equivalentideas. This is a big improvement in usability for the tables!

I agree that on small screens the number of rows displayed should be cut down. Please put in a separate ticket for that.

The problem with making javascript update the number of displayed rows is that it ends up depending  on having too intimate a knowledge of what's displayed on the screen and so you make any change to how things are displayed and it's all too easy to break that.

A better, more maintainable way of doing this would be to selectively display different content for different screen sizes using the normal bootstrap way of hiding content. So, there would be two versions of the number of rows displayed but with one of them hidden at any time. With the rows of data  beyond the rows we would display on small screens we would surround them with a block which gets hidden on small screen sizes.

PR #602, opened 2015-04-17
Incorporate the flash/alert into the head on Owners#show and homepage

### LINKED ISSUE #548: Flash messages appears clunkily above other content
morph mlandauer

equivalentideas:
This adds a wrapping element around the alert, which is then styled specially on the Owners#show page and the home page.

This is because currently it looks odd/broken, jutting into the header on these pages. See #548 .

I've added a class to the body element to indicate the controller and action, as we have on They Vote For You, which allows us to make these style changes.

fixes #548

screen shot 2015-04-17 at 5 36 01 pm
screen shot 2015-04-17 at 5 36 09 pm
screen shot 2015-04-17 at 5 37 06 pm
screen shot 2015-04-17 at 5 37 11 pm

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -5,6 +5,7 @@
 @import "bootstrap";
 @import "font-awesome";
 @import "global/site-header";
+@import "global/alerts";
 @import "bootstrap_and_overrides";
 @import "documentation";
 @import "scrapers";
--- a/app/assets/stylesheets/global/_alerts.scss
+++ b/app/assets/stylesheets/global/_alerts.scss
@@ -1,0 +1,5 @@
+.alert-wrapper {
+  body.owners.show & {
+    background-color: $offset-color;
+  }
+}
--- a/app/assets/stylesheets/global/_alerts.scss
+++ b/app/assets/stylesheets/global/_alerts.scss
@@ -2,4 +2,12 @@
   body.owners.show & {
     background-color: $offset-color;
   }
+
+  body.static.index & {
+    position: absolute;
+    top: 3.4em;
+    background: transparent;
+    z-index: 99;
+    width: 100%;
+  }
 }
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -18,4 +18,8 @@ module ApplicationHelper
   def bs_nav_link(text, url)
     content_tag(:li, link_to(text, url), class: ("active" if current_page?(url)))
   end
+
+  def body_class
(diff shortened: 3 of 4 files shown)


mlandauer:
@equivalentideas from chat earlier

> @matthew do you know an easy way to test if you’re on a specific controller and action?

> You can do it but it’s very likely a “code smell” - as in a sign that something is wrong in the code design
> I think it would be better to be in the same place at least for the foreseeable future
> Every decision like that makes things harder to maintain and they can really add up over time
> So there has to be a really really strong reason for doing it

Have styling that's dependent on the controller and action is an anti-pattern and I don't want to be resorting to it unless it's *absolutely* necessary. So, I won't merge this PR in its current form.

equivalentideas:
> I think it would be better to be in the same place at least for the foreseeable future ...
> Every decision like that makes things harder to maintain and they can really add up over time
So there has to be a really really strong reason for doing it ... Have styling that's dependent on the controller and action is an anti-pattern and I don't want to be resorting to it unless it's absolutely necessary. So, I won't merge this PR in its current form.

That makes sense, I think I was trying too hard not to change the alerts basic appearance. I think a better solution is to make the alert full width everywhere, like on [They Vote For You](https://theyvoteforyou.org.au/), which suits our full width headers style to fix #548:

screen shot 2015-04-20 at 11 59 46 am

mlandauer:
@equivalentideas what i found difficult here is that i explained why making layout dependent on controller/action is not a good thing to do. Then, the next thing I new you submitted a PR doing exactly that.

That made me feel frustrated and essentially ignored.

mlandauer:
If a full width flash can be used consistently everywhere then that's cool by me

equivalentideas:
> @equivalentideas what i found difficult here is that i explained why making layout dependent on controller/action is not a good thing to do. Then, the next thing I knew you submitted a PR doing exactly that.

> That made me feel frustrated and essentially ignored.

I'm sorry about that, I just got a bit confused. I was thinking about moving the markup based on the controller/action and for some reason it didn't connect that doing the same thing on the presentation layer would be a problem. Then I went and changed the styles for those pages when I should have been thinking about changing the design to suit the whole site consistently. A Friday arvo brain error.

mlandauer:
Thanks for the honest explanation @equivalentideas. I appreciate it. It makes me feel better as well.

mlandauer:
Next time we talk I'll try to explain a little more of the "why" making a view or action dependent on the controller/action makes things brittle and harder to maintain

equivalentideas:
That sounds great

PR #616, opened 2015-04-20
[For Review] Provide some basic instructions for new scrapers

### LINKED ISSUE #330: On new scraper gives help on the scraper page
When someone has created a new scraper they are given a scraper page which is blank and it's not obvious what to do next unless you're already familiar with morph.io . So it would be really good if a new scraper had some big text at the top that explained what to do next. e.g. "clone the repo or run the scraper". Right now it doesn't do very much if you run a new scraper so it's probably all a bit confusing

equivalentideas:
If the scraper hasn't been run, give some instructions for cloning
or editing the scraper file on github. Also give a prompt to
run this scraper.

This is a basic first pass modeled on github's new repo instructions. This could be improved on style wise and we could give special instructions if it's a fork from github or scraperwiki.

screen shot 2015-04-20 at 4 18 33 pm

Github new repo page:
screen shot 2015-04-20 at 4 01 43 pm

closes #330

### DIFF
--- a/app/views/scrapers/_show_normal.html.haml
+++ b/app/views/scrapers/_show_normal.html.haml
@@ -38,6 +38,26 @@
 %hr
 .row
   .col-md-10
+    - unless scraper.last_run
+      .panel.panel-default
+        .panel-heading
+          %h2.panel-title Get started
+        .panel-body
+          %h3 Write or edit your new scraper
+          %p
+            Write or edit your
+            = link_to "#{scraper.main_scraper_filename} file directly on github", scraper.github_url_main_scraper_file
+            or on your local machine by cloning the repo with:
+          %p
+            %code git clone #{scraper.git_url_https}
+
+          - if scraper.runnable?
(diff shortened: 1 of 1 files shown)


equivalentideas:
@mlandauer assigned this to you after our discussion that this needs more thinking.

mlandauer:
I'm going to close this PR for the time being. I'm not sure this is the right approach to solving the problem as it adds more complexity to what the user is faced with. It would be good to come up with a solution where the natural state of the UI leads the user down a clear path without having to say "push this button. then do that"

It reminds of what Jef Raskins said in I think the [Humane Interface](http://www.amazon.com/The-Humane-Interface-Directions-Interactive/dp/0201379376) about when people put signs on doors as to whether you have to push or pull is a sign that the user interface of the door is giving the wrong signals.

So, at some later stage we should back to this and think deeper about it.

equivalentideas:
Sounds good :+1:

PR #634, opened 2015-04-22
Search results pagination

### LINKED ISSUE #632: Pagination of search results?
Should we paginate the results?

equivalentideas:
First pass of search results pagination. Set to 10 results per page, which seems like a good amount to scan quickly then click through.

closes #632 

screen shot 2015-04-22 at 3 22 35 pm


mlandauer:
Looking at this now

Issue #507, opened 2015-03-30
Add search over people


mlandauer:
By handle / name. Related to #103

equivalentideas:
@mlandauer are you free this arvo for a real quick chat in slack about this to see if I'm on the right track with this? I'm not sure if I'm thinking in too small iterations.

Story: As a user, when I want to find a person I know on morph, I want to see them in the results when I search their name.

*Louise searches Longair using the header search form looking for Mark Longair.*

How does Louise distinguish between scrapers and people in the results?

I think on the results page we have a big search form with the term entered, so the person can quickly iterate through searches. This is a familiar pattern from places like [github](https://github.com/search?utf8=%E2%9C%93&q=testing), [google](https://www.google.com.au/search?q=testing), [stackoverflow](http://stackoverflow.com/search?q=testing) etc.

Here's a few rough ideas for consideration for how we incorporate 'people' into the results:

1. **People and Scrapers both appear in the same list of results sorted by relevance** 
![2015-04-21 13 33 52](https://cloud.githubusercontent.com/assets/1239550/7245332/9f212ca4-e82c-11e4-9869-d58a54bd7dea.jpg)
2. **People are Scrapers are in separate lists on the page**
![2015-04-21 13 33 43](https://cloud.githubusercontent.com/assets/1239550/7245330/9d0baa66-e82c-11e4-9213-fdfddc1030fa.jpg)
Similar to [They Vote For You](https://theyvoteforyou.org.au/search?utf8=%E2%9C%93&query=Bishop).
3. **Results for Scrapers and people have their own separate pages/tab panes**
![2015-04-21 13 33 30](https://cloud.githubusercontent.com/assets/1239550/7245325/796637d4-e82c-11e4-9d9a-cdf0c238b6ae.jpg)

I think the 3rd approach, Results for Scrapers and people have their own pages/tab panes, will be clearest of these approaches for someone looking for people and will keep the results clear for people searching for scrapers and data. It's also similar to github's approach to search.

Approach 1 with the single list makes navigation simple and, if the results are good, could make scanning and finding what you're after very quick. If there's lots of similarly relevant results it could get really messy for people trying to scan through different types of information. Also if you're looking for just scrapers, or just people, there are lots of unhelpful results.

I think Approach 2 is confusing as you're not sure if there are more results below if the first list is very long.

## Displaying people/organisation items

[...]

PR #640, opened 2015-04-22
make the user results more human and scanable

### LINKED ISSUE #633: The display of user search results is messy and should do more to emphasise the human.
Here's the current display of user search results:

screen shot 2015-04-21 at 1 25 24 pm

Here's the discussion of this on #507:

> In the results for people I think we prioritise identifying information, rather than information about their morph activity. So for people: full name, image, user name, organisations, above number of scrapers, amount of data collected. For organisations: full name, image, user name, number of members (maybe member images?), above number of scrapers etc.

equivalentideas:
First pass of displaying users in search results. This simplifies the display by stripping out unnecessary information and decoration and giving the information some hierarchy. It makes the user's avatar slightly larger and puts the focus on their full name to emphasise the human.

If we want we could apply this style to the /users page as well by simply removing the 'if @q' logic in the _user.html.haml and _organisation.html.haml partials.

closes #633 

In a next iteration we might want to have 2 or 3 results per row, to get more to scan on the screen and use more of the available screen space. If we do that we might want to bring back the borders.

## Before
screen shot 2015-04-22 at 6 42 19 pm
screen shot 2015-04-22 at 6 42 40 pm

## After
screen shot 2015-04-22 at 6 36 21 pm
screen shot 2015-04-22 at 6 36 30 pm
screen shot 2015-04-22 at 6 40 18 pm
screen shot 2015-04-22 at 6 40 26 pm
screen shot 2015-04-22 at 6 40 45 pm

### DIFF
--- a/app/assets/stylesheets/users.css.scss
+++ b/app/assets/stylesheets/users.css.scss
@@ -38,4 +38,10 @@
     margin-top: 5px;
     color: $link-color;
   }
+
+  .owner-nickname {
+    margin-bottom: 5px;
+    display: block;
+    font-size: $font-size-large;
+  }
 }
--- a/app/assets/stylesheets/users.css.scss
+++ b/app/assets/stylesheets/users.css.scss
@@ -24,6 +24,16 @@
   display: block;
   color: $text-color;
 
+  &:hover,
+  &:focus {
+    text-decoration: none;
+
+    .media-heading {
+      color: $link-hover-color;
+      text-decoration: $link-hover-decoration;
+    }
+  }
+
   + .owner-item {
     margin-top: 1.5em;
   }
--- a/app/assets/stylesheets/users.css.scss
+++ b/app/assets/stylesheets/users.css.scss
@@ -22,17 +22,6 @@
 
 .owner-item {
   display: block;
-  color: $text-color;
-
-  &:hover,
-  &:focus {
-    text-decoration: none;
-
-    .media-heading {
-      color: $link-hover-color;
-      text-decoration: $link-hover-decoration;
-    }
-  }
 
   + .owner-item {
     margin-top: 1.5em;
--- a/app/assets/stylesheets/users.css.scss
+++ b/app/assets/stylesheets/users.css.scss
@@ -34,6 +34,7 @@
   .media-heading {
     margin-top: 3px;
     color: $link-color;
+    font-weight: bold;
   }
 
   .owner-nickname {
--- a/app/assets/stylesheets/users.css.scss
+++ b/app/assets/stylesheets/users.css.scss
@@ -22,6 +22,7 @@
 
 .owner-item {
   display: block;
+  color: $text-color;
 
   + .owner-item {
     margin-top: 1.5em;
@@ -35,5 +36,6 @@
 
   .media-heading {
     margin-top: 5px;
+    color: $link-color;
   }
 }
--- a/app/assets/stylesheets/users.css.scss
+++ b/app/assets/stylesheets/users.css.scss
@@ -32,7 +32,7 @@
   }
 
   .media-heading {
-    margin-top: 5px;
+    margin-top: 3px;
     color: $link-color;
   }
 
--- a/app/assets/stylesheets/users.css.scss
+++ b/app/assets/stylesheets/users.css.scss
@@ -21,8 +21,6 @@
 }
 
 .owner-item {
-  display: block;
-
   + .owner-item {
     margin-top: 1.5em;
   }
--- a/app/assets/stylesheets/users.css.scss
+++ b/app/assets/stylesheets/users.css.scss
@@ -19,3 +19,11 @@
   padding-top: 15px;
   padding-bottom: 15px;
 }
+
+.owner-item {
+  display: block;
+
+  + .owner-item {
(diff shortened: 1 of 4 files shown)


henare:
A few thoughts:

My first reaction was that these SEEM REALLY HUGE :) They feel a little out of place, size-wise, with the rest of the page (although now I look at it the search box looks in proportion to these things and also a little out of proportion to the rest of the page).

When hovering over an item all the text changes colour but only the title part gets underlined. I think we need to either make everything get underlined or make only the title change colour.

I think we should remove the URLs being displayed on some profiles. I tried to click on the URLs in peoples profiles and expected it to go to that address. URLs are also visually noisy anyway. 

>If we want we could apply this style to the /users page as well by simply removing the 'if @q' logic in the _user.html.haml and _organisation.html.haml partials.

I think that's a good idea to keep things consistent.

The existing `/users` page also has how many scrapers that Owner has which I think is useful because it shows you before you click if there's any useful content on that person's page. I think it should be brought back into this redesign. Was it removed for a reason?

equivalentideas:
> The existing /users page also has how many scrapers that Owner has which I think is useful because it shows you before you click if there's any useful content on that person's page. I think it should be brought back into this redesign. Was it removed for a reason?

This is from the issue #633 :

> In the results for people I think we prioritise identifying information, rather than information about their morph activity. So for people: full name, image, user name, organisations, above number of scrapers, amount of data collected.

> I think the simple first step here is to adjust the hierarchy/sizes of the personal information and remove the scraper count. I think there is room for doing something interesting with activity information, to reward and highlight people doing great work, but i think that is for a further iteration.

I see what you're saying about it being useful if you don't know the people already. This first iteration is mainly supporting the use case of searching for people who you know ( some stuff about that in https://github.com/openaustralia/morph/issues/507#issuecomment-94633432 ) .

I think for this first iteration we remove the scraper activity information and consider it as a new issue/use story.

What do you think @henare ? If we keep it out, does that effect what you think about implementing this style on the /users page?

equivalentideas:
> I think we should remove the URLs being displayed on some profiles. I tried to click on the URLs in peoples profiles and expected it to go to that address. URLs are also visually noisy anyway.

I think the urls are useful in recognising the user, and are interesting information to morph users who are developers—but I agree that it's definitely confusing to not be able to click on them.

If we keep them we should make them clickable, and restrict the main link to the heading, name and image.

> URLs are also visually noisy anyway.

If we keep em we could strip off the https://www. bit and just show the domain, which is the recognisable bit like on twitter profiles:

screen shot 2015-04-27 at 5 00 11 pm

equivalentideas:
Re the URLS, I recon we drop them off as keeping them is more complex and this is supposed to be a quick first iteration :)

equivalentideas:
I just accidentally clicked on a user but clicking the whitespace to the right (the whole block element is the link). I think we should restrict the link to image and main heading as github do https://github.com/search?q=henare&type=Users&utf8=%E2%9C%93

henare:
>I think for this first iteration we remove the scraper activity information and consider it as a new issue/use story.
>
>What do you think @henare ? If we keep it out, does that effect what you think about implementing this style on the /users page?

If this is a first iteration and it will be brought back later it doesn't really matter. I think consistency is more important in this case so we should apply the style to the /users page. Applying it more widely as we're iterating on it will allow us to see the problems quicker too.

>Re the URLS, on second thoughts, I recon we drop them off as keeping them is more complex and this is supposed to be a quick first iteration :)

:+1: 

>I just accidentally clicked on a user but clicking the whitespace to the right (the whole block element is the link). I think we should restrict the link to image and main heading as github do https://github.com/search?q=henare&type=Users&utf8=%E2%9C%93

:+1:

equivalentideas:
> If this is a first iteration and it will be brought back later it doesn't really matter. I think consistency is more important in this case so we should apply the style to the /users page. Applying it more widely as we're iterating on it will allow us to see the problems quicker too.

Great, thanks @henare .

So the changes to makes are:

* [ ] restrict the link to the image and heading
* [ ] remove urls
* [ ] apply this to the /users page (all user lists)

> My first reaction was that these SEEM REALLY HUGE :) They feel a little out of place, size-wise, with the rest of the page (although now I look at it the search box looks in proportion to these things and also a little out of proportion to the rest of the page).

I'll try it out a little smaller. We're trying to really emphasise the person with these big images, but maybe I've gone a bit over board ;)

equivalentideas:
I've made those changes and ended up reducing the size of the search input and the user items. I think @henare is right and it was a bit out of step with the current site typography/hierarchy.

screen shot 2015-04-27 at 10 12 34 pm

screen shot 2015-04-27 at 10 14 10 pm

mlandauer:
The way that users and scrapers are now shown in a list is inconsistent. Each item in the list should be separated by a line and the whole item should be a link.

mlandauer:
@henare I see you've merged this already. I'm going to revert it until the issue above is dealt with.

henare:
I've just aborted the deploy I was doing so production is still at 9cef90e.

mlandauer:
@henare thanks!

henare:
I was hoping I could reopen this but it doesn't look like it works that way so I've opened #649, assigned to @equivalentideas.

PR #651, opened 2015-04-28
show scraper language on search results

### LINKED ISSUE #639: Show language of scraper on search results page

equivalentideas:
This adds the scraper language to the scraper element in search results, so citizens can quickly find scrapers they can work with in a language they can understand, when they are searching for scrapers.

This doesn't effect sync version of the scraper partial, used on the /scrapers page or owner show page, just the static version used in search results. Do we want to apply it on the sync version for consistency?

closes #639 

The style the language is faded, not the most salient element in the block, but clear on the righthand side of the scraper-item. This draws on Githubs style for showing the language of repos.

screen shot 2015-04-28 at 10 50 06 am
screen shot 2015-04-28 at 10 50 22 am

### DIFF
--- a/app/assets/stylesheets/scrapers.css.scss
+++ b/app/assets/stylesheets/scrapers.css.scss
@@ -216,10 +216,17 @@ a i.fa-clock-o {
     margin: .2em 0 0 .25em;
     color: lighten($gray-light, 10%);
     font-weight: bold;
+    width: 4em;
+    text-align: right;
   }
 
   .icon-box {
     margin: 0 .25em;
+    text-align: center;
+
+    @media (min-width: $grid-float-breakpoint) {
+      width: 3em;
+    }
 
     .spinner,
     i {
--- a/app/assets/stylesheets/scrapers.css.scss
+++ b/app/assets/stylesheets/scrapers.css.scss
@@ -213,10 +213,22 @@ a i.fa-clock-o {
   }
 
   .scraper-lang {
-    margin-top: .2em;
-    padding-left: .5em;
+    margin: .2em 0 0 .25em;
     color: lighten($gray-light, 10%);
     font-weight: bold;
+  }
+
+  .icon-box {
+    margin: 0 .25em;
+
+    .spinner,
+    i {
+     vertical-align: text-top;
+   }
+
+    + .icon-box {
+      margin: 0 .25em 0 .5em;
+    }
   }
 }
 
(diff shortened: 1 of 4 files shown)


mlandauer:
That looks good

Yes, we should apply it on the sync version for consistency as well.

Note that when you update the partial `app/views/sync/_scraper.html` there's a "cache" block which basically says cache everything inside this block. Whenever you change anything inside of a cache block you need to update the version.

So, 
```
- cache ["v2", scraper] do
```
will need to change to
```
- cache ["v3", scraper] do
```

Also, note that the call to `Scraper#language` is slow because it's accessing the file system for each scraper to figure out the language. Let's not worry about that for the time being but have it in the back of our minds in case it does feel slow.

Also please cherry-pick https://github.com/openaustralia/morph/commit/637bc6521ed1074325d160c0230a9cd235b8b81b on to this branch so the travis ci tests will run correctly

equivalentideas:
Thanks for that @mlandauer , I'll make those changes.

equivalentideas:
I've added the lang to the sync version. Had to do a bit of finessing of the spacing with the icons across screen sizes.

screen shot 2015-04-28 at 12 15 22 pm
screen shot 2015-04-28 at 12 15 43 pm

mlandauer:
It looks wrong to me that the fixed width icons are now not in a fixed position. Makes it harder to scan down the page and makes those icons less useful. 

How about putting the fixed width icons below the language?

equivalentideas:
I see what you mean @mlandauer . Here's what they look like below:

screen shot 2015-04-28 at 2 15 32 pm

And here's what it looks like with each item having a set width on wide screens:

screen shot 2015-04-28 at 2 09 18 pm

I think in either case it would be a good idea to swap the order of the running icon and the clock, because the clock is much more common and it will make it easier to see if the 'running' icon is there if it's not in the middle.

I think the second option where they're all on the same line is simpler/clearer.

mlandauer:
I'm with you, having them all on one line is clearer. Leave the order of the icons unchanged. If that happens that needs to be done consistently everywhere and be done separately from this PR.

equivalentideas:
> I'm with you, having them all on one line is clearer. Leave the order of the icons unchanged. If that happens that needs to be done consistently everywhere and be done separately from this PR.

On it.

equivalentideas:
@mlandauer I've done that and added a fix for when there's so language defined so the icons still stay inline.

I'm having trouble with my dev version getting scrapers running. I've been copying markup from the live site to test this, I'm pretty confident it's solid, but it would be great if you would give it a quick test as well.

Issue #319, opened 2014-04-24
Sort scrapers in error email newest error to oldest


henare:
Because you want to know about new problems first, e.g. there's a scraper you know if problematic but will take ages to fix so you put it off but you want to know if there's a new error on a different scraper because it could be an easy fix.

PR #652, opened 2015-04-28
Reorder alert email

henare:
This is what happens when you try to be smart and refactor things *before* doing the dumb thing and just fixing what you set out to do.

I'll open a PR with just the fix for #319 and this can be closed.


henare:
Closing in favour of #653.

PR #653, opened 2015-04-28
Reverse order of scrapers in alert email. Fixes #319

### LINKED ISSUE #319: Sort scrapers in error email newest error to oldest

henare:
A much simpler version of #652.


henare:
Just got my first reversed email - handy! :smile:

PR #657, opened 2015-04-28
When scraper is manually stopped show it not as an error

### LINKED ISSUE #626: When scraper is manually stopped show it not as an error
Currently it will say that the scraper has errored with some error code.

It would be much better if it said something like
"Last run manually stopped 3 minutes ago"

henare:
selection_001
selection_002

Fixes #626

### DIFF
--- a/app/models/run.rb
+++ b/app/models/run.rb
@@ -69,6 +69,10 @@ class Run < ActiveRecord::Base
     finished? && !finished_successfully?
   end
 
+  def manually_stopped?
+    finished? && status_code == 130
+  end
+
   def error_text
     log_lines.where(stream: "stderr").order(:number).map{|l| l.text}.join
   end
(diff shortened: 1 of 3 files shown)


mlandauer:
I'm not sure, just by looking at the code, that this is going to do everything we want it to do. Will this scraper still be showing as errored in lists of scrapers? My guess is it will.

I think the hardcoded error code to represent a manual stop was a hack and this is a hack on top of hack ;-)

Is there a cleaner way?

henare:
>I think the hardcoded error code to represent a manual stop was a hack and this is a hack on top of hack ;-)

Fair call. The geek in me really likes the Ctrl-C error code :)

>Is there a cleaner way?

Since this needs a bit of a rethink I'll close this PR for the moment.

Issue #595, opened 2015-04-15
Show users who downloaded data / used the api for a scraper


mlandauer:
Related to #122

This gives users a sense of who and how many people are using the data they've scraped. Also by adding visibility we potentially foster more interaction between users and hopefully more sense of community and more scraping...

equivalentideas:
As morph.io is the scraper platform for civic hackers, are there privacy concerns here for activists or journalists using morph.io having their research activity made public and searchable?

I feel like this is an issue, but haven't seen any cases on point. Just feels :S

If there are privacy issues, we don't have to make this completely public to get the benefits above. We could just show Scraper Owners or Logged in users. We could also phase this in and see how it goes.

equivalentideas:
Here's a few examples of how this is done on some other platforms:

## Freesound.org

Repo of open field recordings. Shows count and people who've downloaded a [recording page](http://www.freesound.org/people/lukebacon/sounds/169134/)

![screen shot 2015-04-29 at 11 10 54 am](https://cloud.githubusercontent.com/assets/1239550/7383714/897849e6-ee67-11e4-80f3-f8654514ca8a.png)

and the user names of people who've downloaded on a [recording/downloaders page](http://www.freesound.org/people/lukebacon/sounds/169134/downloaders/), it separated by day:

![screen shot 2015-04-29 at 12 00 36 pm](https://cloud.githubusercontent.com/assets/1239550/7383709/6e3d0be4-ee67-11e4-964f-5aa740f40aef.png)

## Github

Similar concept, Github shows people who've forked a repo:

![screen shot 2015-04-29 at 11 18 00 am](https://cloud.githubusercontent.com/assets/1239550/7383742/0423ea4c-ee68-11e4-9345-e4fc6a15ecb4.png)

There display of contributors count in the prominently beneath the description is also worth noting:

![screen shot 2015-04-29 at 11 17 11 am](https://cloud.githubusercontent.com/assets/1239550/7383737/fb346736-ee67-11e4-831f-27935684f227.png)

mlandauer:
@equivalentideas I think it's important to think through the privacy implications

I'm keen on trying, as an experiment, to make public who downloads/uses data because it has the potential to slightly redress the imbalance between the people that create/scrape data and publish it for free on morph.io and those that potentially use/abuse it.

More concretely, if I put in the effort to scrape some data and give it away for free, shouldn't I be able to see who downloads it and potentially see what they do with it as well? It seems like a fair bargain to me.

[...]

PR #679, opened 2015-04-30
[WIP] Show scraper usage

equivalentideas:
As a user, I want to know how much, and by who, the data I've scraped is being used, so that I can connect with people with similar interests, discover opportunities, know if there is interest in my work and what people are doing with it.

This adds a display of the *number of downloads of a scraper's data* and *number of people who've downloaded it*, to scraper show pages.

This is a static HTML prototype at the moment.

screen shot 2015-04-30 at 2 29 50 pm
screen shot 2015-04-30 at 2 25 30 pm

### DIFF
--- a/app/models/scraper.rb
+++ b/app/models/scraper.rb
@@ -23,6 +23,8 @@ class Scraper < ActiveRecord::Base
 
   has_one :last_run, -> { order "queued_at DESC" }, class_name: "Run"
 
+  has_many :api_queries
+
   validates :name, format: { with: /\A[a-zA-Z0-9_-]+\z/, message: "can only have letters, numbers, '_' and '-'" }
   validates :owner, presence: true
   validates :name, uniqueness: { scope: :owner, message: 'is already taken on morph.io' }
@@ -42,12 +44,8 @@ class Scraper < ActiveRecord::Base
     (watchers + owner.watchers).uniq
   end
 
-  def downloads
-    ApiQuery.where(scraper_id: id)
-  end
-
   def downloaders
-    downloads.map do |d|
+    api_queries.map do |d|
       Owner.find(d.owner_id)
     end.uniq
   end
--- a/app/models/scraper.rb
+++ b/app/models/scraper.rb
@@ -46,6 +46,12 @@ class Scraper < ActiveRecord::Base
     ApiQuery.where(scraper_id: id)
   end
 
+  def downloaders
+    downloads.map do |d|
+      Owner.find(d.owner_id)
+    end.uniq
+  end
+
   # Given a scraper name on github populates the fields for a morph.io scraper but doesn't save it
   def self.new_from_github(full_name, octokit_client)
     repo = octokit_client.repository(full_name)
--- a/app/models/scraper.rb
+++ b/app/models/scraper.rb
@@ -45,9 +45,7 @@ class Scraper < ActiveRecord::Base
   end
 
   def downloaders
-    api_queries.map do |d|
-      Owner.find(d.owner_id)
-    end.uniq
+    api_queries.group(:owner_id).map {|d| d.owner}
   end
 
   # Given a scraper name on github populates the fields for a morph.io scraper but doesn't save it
--- a/app/models/scraper.rb
+++ b/app/models/scraper.rb
@@ -42,6 +42,10 @@ class Scraper < ActiveRecord::Base
     (watchers + owner.watchers).uniq
   end
 
+  def downloads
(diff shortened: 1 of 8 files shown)


equivalentideas:
What do you think @mlandauer ?

To show who is using the data, I think we could show an avatar or user name instead of the count if there are 1 - 3 users.

If there are more, one thing we could do a disclosure thing where you click '5 users' and an inline list of user avatars/names unfolds on a line below the sentence.

mlandauer:
What to do for a small number of users is an optimisation best left after the list of users is implemented. So, I think put that to one side.

Also, looking at the information in place, starting with a number makes it less clear. I think a slight improvement would be "Downloaded 1934 times by 3 users"

What's the reason for putting the list of users in a disclosure rather than on a separate page?

And will it show the number of downloads for each user in that list?

I think adding a static prototype for that list would be a good thing to do now

equivalentideas:
From #595 :
> One more thing to consider the downloaders can also be organisations. This will happen in the case where an api key is used which belongs to an organisation. This is how PlanningAlerts applications are downloaded for example. So, the design will need to incorporate that in some way

To make it super clear who (especially if its an org) is using the data, I've put in the avatars of the person instead of just the count.

That was before @mlandauer last comment.

> What to do for a small number of users is an optimisation best left after the list of users is implemented. So, I think put that to one side.

You know this data better than I do, do you think a small number is the less common case or it will often be lots of downloaders?

> What's the reason for putting the list of users in a disclosure rather than on a separate page?

I think this is dependent on whether there are often lots of downloaders or a small number.

* If the majority case is many downloaders, that seems like something we'd move away so it can be viewed simply in it's own space.
* If it's a small number of downloaders, then I think it's simpler to keep it on the page next to the data and show it inline, no disclosure, instead of the count.

Again, you know the usage better than I do, so I'm gonna go with what you're suggesting and we can iterate.

> And will it show the number of downloads for each user in that list?

I wasn't thinking so for a first iteration, but if it supports a use case let's do it.

equivalentideas:
Here's the latest version of this with a static list of contributors:

screen shot 2015-04-30 at 4 25 16 pm

> And will it show the number of downloads for each user in that list?

I can see how this is particularly useful to the scraper author. I'm going to see what this is like added to the tool tip.

Also, we should could order the list by number of downloads.

equivalentideas:
> I can see how this is particularly useful to the scraper author. I'm going to see what this is like added to the tool tip.

Here's what this is like:
screen shot 2015-04-30 at 4 40 43 pm

I think that works quite well, but it depends how important that data is for most users—if it's crucial to understand the use, lets try a different approach where hover isn't required.

equivalentideas:
I've managed to get this working dynamically. I bet it could be done more elegantly and performant.

screen shot 2015-05-01 at 4 46 42 pm

Currently if there are no dowloads it says "Downloaded 0 times". That could be a kick to do something with the data ;-)

I forgot to do the pluralize on times—will add that now.

mlandauer:
There's definitely a big performance issue with this and some refactoring that could be done. It's easily solved. I can either do it on this branch, or I can comment on the code and let you do it.

Which will it be?

app/models/scraper.rb
```diff
@@ -42,6 +42,16 @@ def all_watchers
     (watchers + owner.watchers).uniq
   end
 
+  def downloads
```
mlandauer: You should do this instead with an association on this model

app/models/scraper.rb
```diff
@@ -42,6 +42,16 @@ def all_watchers
...
+    ApiQuery.where(scraper_id: id)
+  end
+
+  def downloaders
+    downloads.map do |d|
```
mlandauer: It will super memory hungry and slow if all the `api_queries` get loaded into memory. Really, you want to use something like

```
downloads.group(:owner_id)
```

which will only return records with different owners. So, you can drop call to `uniq` as well.

Also, to simplify the creation of owner objects add an association on the `ApiQuery` class so that you can just call `owner` on an `ApiQuery` object and get the owner.

mlandauer:
I've pushed the feature flag stuff to this PR

app/models/scraper.rb
equivalentideas: would this bit go into the controller instead of the model?

mlandauer: Are you talking about the `downloaders` method? If so, keep that on the model.

equivalentideas: Great thanks.

Issue #589, opened 2015-04-15
Python buildpack is not stopping on erroring


mlandauer:

mlandauer:
I've seen this issue with Python. Trying to run a ruby scraper without a Gemfile.lock stops nicely during the compile phase

mlandauer:
I can reproduce the problem with python if I add a `requirements.txt` file with `Pygments==1.2.3.4` in it.

mlandauer:
Upon some further investigation this is looking like an issue either with buildstep or the python buildpack

mlandauer:
I don't think this needs to be a blocker to making buildstep the default way of doing things. So, removing the milestone.

mlandauer:
A very likely fix is in a later version of python buildpack https://github.com/heroku/heroku-buildpack-python/commit/8f4eb44d36efbb372b3834f1d69246b3001d934b

PR #682, opened 2015-04-30
Use the same libraries as ScraperWiki had in our repo templates

### LINKED ISSUE #681: New repositories don't contain useful libraries
Prior to default buildpacks, if you created a scraper you got a bunch of useful libraries. Now you get a very minimal set. For example, the default Gemfile only includes the scraperwiki gem.

This was noticed by @woodbine. They created a scraper and expected to be able to use BeautifulSoup, which is in the default `requirements.txt` but commented out.

In that example, just uncommenting it isn't the solution. Going back to the Ruby example, our `scraper.rb` template includes a reference to Mechanize but that's not included in our default Gemfile!

henare:
When you create a new repo you expect it to be useful straight away.
This change makes it just like it was before buildpack support but of
course you can now change these libaries to suit you.

Fixes #681


henare:
Admittedly a bit of a knee jerk reaction so opening for review by @mlandauer.

mlandauer:
I think there are a few different approaches to how to include the configuration files (like `Gemfile`) in the buildpacks template:

1. We don't include the configuration files at all - so it will by default use the scraperwiki setup
2. We include the scraperwiki configuration as default - as is suggested in this PR
3. We include a stripped down simple configuration

Advantages and disadvantages of each:

1. We don't include the configuration files at all
  * Advantage: Very simple for the user at first. No extra files. Also, closest to the way that morph.io (and scraperwiki classic) used to work.
  * Disadvantage: Once the user wants to add a new library it's harder to figure out what to do

2. We include the scraperwiki configuration as default
  * Advantage: Makes it very clear to the experienced user what is going on and what libraries are included by default. Also, a new template scraper has the same libraries by default as are available by default in the non-buildpacks setup.
  * Disadvantage: Noisy. You end up with long list in a `Gemfile` or `requirements.txt` which might be overwhelming to a newcomer. What are all these libraries? What has this got to do with my scraper?

3. We include a stripped down simple configuration
   * Advantage: Installs just the libraries needed to run the examples. Makes it clearer what to do next if you want to install new libraries. (We could include a link to the morph.io documentation and an explanation in the config file - See https://github.com/openaustralia/morph/issues/683). Also we can install new versions of the necessary libraries by default rather than being stuck with the old stuff that ScraperWiki used. It allows us to unchain ourselves from the past
  * Disadvantage: Behaves differently than the old way on morph.io and breaks more with the past

Option 1 is what we used to do. Option 2 is what's being suggested in this PR.

I think option 3 is preferable. What would need to be done to make that option work for the issue that's being addressed in this PR?

henare:
Another idea I had is to include `Gemfile-example`-type files for each language's template that contain comments on how to get them working.

However the more I think about it the more I think option 3 is best too.

>What would need to be done to make that option work for the issue that's being addressed in this PR?

The issue we're fixing is "New repositories don't contain useful libraries" and if we "Install just the libraries needed to run the examples" then I think we're providing enough useful libraries to get started, e.g. the `scraperwiki` and `mechanize` libraries for Ruby.

mlandauer:
What about the issue you fixed last night with the python library version.
Was that related to https://github.com/openaustralia/morph/issues/589?

Would that have been helped by doing
https://github.com/openaustralia/morph/issues/683?

henare:
>What about the issue you fixed last night with the python library version.
>Was that related to https://github.com/openaustralia/morph/issues/589?
>
>Would that have been helped by doing
>https://github.com/openaustralia/morph/issues/683?

I don't think these are related to [what I did last night](https://github.com/woodbine/sp_E3520_SCC_gov/pull/1/files).

The scraper that had a problem last night was trying to use BeautifulSoup 4. In our Python template there's a commented out line for BS but that had not been uncommented. Even if they had uncommented it, the library name was wrong for that version.

mlandauer:
Did you update morph.io's python template to fix the issue?

henare:
>Did you update morph.io's python template to fix the issue?

No, I didn't change anything on morph.io.

mlandauer:
Will you or shall I?

henare:
I think we need to decide on an approach to this. Whatever we choose here will also fix that issue.

henare:
If you agree with option 3 I'm happy to pull together a pull request for that.

mlandauer:
I like option 3 too. So, let's go with that. A PR would be most awesome. Thank you! :-)

henare:
Excellent. I'll close this and work on that.

Issue #683, opened 2015-05-01
Add explanation of how to add new libraries in comments of config template files


mlandauer:
So, for instance, when you create a new Ruby scraper it will include a template `Gemfile` which should include a comment that explains roughly what to do to add new libraries. This should also include a link to the morph.io documentation for that language.

PR #697, opened 2015-05-04
Install just the libraries needed to run the examples

### LINKED ISSUE #681: New repositories don't contain useful libraries

henare:
Fixes #681 by implementing option 3 here: https://github.com/openaustralia/morph/pull/682#issuecomment-98014655

Also included are fixes made along the way to ensure the examples run when simply uncommented.

And a fix for #683 since it seemed sensible to do that at the same time.

### DIFF
--- a/default_files/perl/config/cpanfile
+++ b/default_files/perl/config/cpanfile
@@ -1,3 +1,7 @@
+# It's easy to add more libraries or choose different versions. Any libraries
+# specified here will be installed and made available to your morph.io scraper.
+# Find out more: https://morph.io/documentation/perl
+
 requires "DBD::SQLite", "1.42";
 requires "DBI";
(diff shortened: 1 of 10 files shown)


mlandauer:
Going to look at this next. Sorry for the delay with this @henare!

henare:
Thanks :+1:

Issue #698, opened 2015-05-05
The logged in landing page can do more to help citizens accomplish tasks


equivalentideas:
Currently when logged in, the [home page](https://morph.io/) doesn't do much to help citizens do the things they've come to morph.io for. It was originally put together to help the developers see what was happening in the app. It has lists of new users, active scrapers, and organisations on the site:

![screen shot 2015-05-05 at 10 54 56 am](https://cloud.githubusercontent.com/assets/1239550/7465562/89676614-f315-11e4-8077-e0f13685a010.png)

We can break down the people who visit this page into two groups:

* **Current users**: people who own and watch scrapers on morph
* **New users**: people who don't have scrapers, and likely have just signed up

Here are our current assumptions about the jobs these people want the landing page to do for them:

## Current users

*As a current user, when I log into morph.io, I want to know about scrapers that need my attention, so that I can fix them and keep getting the data I need.*

## New users

*As a new user, when I sign up, I want to discover scrapers that are interesting to me, so I can better understand why scraping is useful and something I should learn more about.*

*As a new user, when I visit the landing page, I want some help to write my first scraper on morph.io, so that I can get the data I need.*

equivalentideas:
@mlandauer and I's initial thoughts about how we can do better for these people:

> As a new user, when I sign up, I want to discover scrapers that are interesting to me, so I can better understand why scraping is useful and something I should learn more about.

The best way we can currently help people find interesting scrapers is with search.

We could add a prominent search banner with some suggested search terms to get people started.

We should also redirect new signups to this landing page, rather than their profile as we currently do.

> As a new user, when I visit the landing page, I want some help to write my first scraper on morph.io, so that I can get the data I need.

We don't really have any documentation of how to write a scraper. We depend on people already knowing how or being skilled enough to work it out by looking at the templates/other scrapers.

We're not planning to add any new how-to's at this point, *but* we could ease people into adding a scraper, and hopefully with determination will prevail.

We had an idea that you could add a scraper from the landing page if you had none—by entering a description and title for it. We could give some direction for a good description and require it. That way, even if people can't follow through and write the scraper, they can make a repo to describe what they're after. This is a good first step for them, would be very interesting experiment, and there may be a way to then connect people who are interested to pick up those scrapers and write them.

> As a current user, when I log into morph.io, I want to know about scrapers that need my attention, so that I can fix them and keep getting the data I need.

[...]

PR #702, opened 2015-05-05
[WIP] Search banner for signed in landing page

### LINKED ISSUE #701: Add search banner to signed in landing page
extracted from #698 

**As a new user, when I sign up, I want to discover scrapers that are interesting to me, so I can better understand why scraping is useful and something I should learn more about.**

The best way we can currently help people find interesting scrapers is with search.

* [ ] Add a prominent search banner with some suggested search terms to get people started.
* [ ] Redirect new signups to this landing page, rather than their profile as we currently do.

Here's an idea for what the search banner form could be like:

Here's how this could sit in a banner on the page:

Maybe thinner, and maybe align the form more to the left of the page rather than centre.

It could be in our general banner style. Maybe a lighter or no background colour.

@mlandauer :
> I think some of those search examples should highlight scrapers for some of the non-english speaking countries where there are a lot of morph.io users (e.g. germany, spain) - it would also be good to include non-european examples I think.

equivalentideas:
To help citizens explore morph and discover scrapers of interest to them, these commits add a simple search form banner sections to the logged in landing page:

screen shot 2015-05-06 at 9 15 54 am
screen shot 2015-05-06 at 9 15 48 am

* [x] Before this is merged we need to settle on the suggested search terms (see [discussion on #701](https://github.com/openaustralia/morph/issues/701#issuecomment-99253000)).

relates to #698 

closes #701

### DIFF
--- a/app/assets/stylesheets/_landing_page.scss
+++ b/app/assets/stylesheets/_landing_page.scss
@@ -12,6 +12,11 @@
 }
 
 .banner-search {
+  @media (min-width: $screen-md-min) {
+    padding-bottom: 24px;
+    padding-top: 24px;
+  }
+
   .label-heading {
     font-size: $font-size-h4;
 
--- a/app/assets/stylesheets/_landing_page.scss
+++ b/app/assets/stylesheets/_landing_page.scss
@@ -31,6 +31,10 @@
     @media (min-width: $screen-md-min) {
       font-size: $font-size-h1;
     }
+
+    .pre-heading {
+      margin-bottom: .2em;
+    }
   }
 }
 
--- a/app/assets/stylesheets/_landing_page.scss
+++ b/app/assets/stylesheets/_landing_page.scss
@@ -11,6 +11,10 @@
   }
 }
 
+.search-suggestions {
+  margin-top: .75em;
+}
+
 .banner-wrapper {
   background-color: #716b58;
 }
--- a/app/assets/stylesheets/_landing_page.scss
+++ b/app/assets/stylesheets/_landing_page.scss
@@ -11,6 +11,24 @@
   }
 }
 
+.banner-search {
+  .label-heading {
+    font-size: $font-size-h4;
+
+    @media (min-width: $screen-xs-min) {
+      font-size: $font-size-h3;
+    }
+
+    @media (min-width: $screen-sm-min) {
+      font-size: $font-size-h2;
+    }
+
+    @media (min-width: $screen-md-min) {
+      font-size: $font-size-h1;
+    }
+  }
+}
+
 .search-suggestions {
   margin-top: .75em;
 }
--- a/app/assets/stylesheets/_landing_page.scss
+++ b/app/assets/stylesheets/_landing_page.scss
@@ -36,6 +36,10 @@
 
 .search-suggestions {
   margin-top: .75em;
+
+  @media (min-width: $screen-md-min) {
+    font-size: 1.125em;
+  }
 }
 
 .banner-wrapper {
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -6,6 +6,7 @@
 @import "font-awesome";
 @import "vendor/tablesaw.stackonly";
 
+@import "global/typography";
 @import "global/site-header";
 @import "global/alerts";
 @import "bootstrap_and_overrides";
--- a/app/assets/stylesheets/global/_typography.scss
+++ b/app/assets/stylesheets/global/_typography.scss
@@ -1,0 +1,3 @@
+.pre-heading {
+  display: block;
+}
--- a/app/assets/stylesheets/search.css.scss
+++ b/app/assets/stylesheets/search.css.scss
@@ -2,7 +2,7 @@
 // They will automatically be included in application.css.
 // You can use Sass (SCSS) here: http://sass-lang.com/
 
-.search-header {
+.search-form {
   margin-bottom: 2em;
 
 }
--- a/app/assets/stylesheets/search.css.scss
+++ b/app/assets/stylesheets/search.css.scss
@@ -5,9 +5,10 @@
 .search-header {
   margin-bottom: 2em;
 
-  + .nav-tabs {
-    margin-bottom: 2em;
-  }
+}
+
+.search-type-nav {
+  margin-bottom: 2em;
 }
 
 .search-results em {
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -23,4 +23,8 @@ module ApplicationHelper
     l = Morph::Language.new(key)
     image_tag(l.image_path, options) + " " + l.human
   end
+
+  def floor_to_hundreds(number)
+    (number / 100.0).floor * 100
+  end
 end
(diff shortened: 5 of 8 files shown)


equivalentideas:
I'm also going to make the '2000 scrapers' dynamic, and ground to 100s, so it'll be 2200 now. That will give return users a nice sense of morph growing as the number increases.

mlandauer:
Looks good.

I think just putting the number of scrapers there dynamically without any fancy rounding is fine.

Also might be nice to emphasise in some way that the scrapers have data which is what you're actually likely to be interested in here.

mlandauer:
Or maybe you could say explicitly

> Search over 2134 scrapers for data you can use

equivalentideas:
> I think just putting the number of scrapers there dynamically without any fancy rounding is fine.

I think here, where we want to get people excited to search, 'over 2200 scrapers', has that slight excitement where '2134 scrapers' feels a bit too accurate ;)

Happy to go with what you think though.

Trying out some variations of the 'for data you can use'.

equivalentideas:
screen shot 2015-05-06 at 10 09 54 am

Here's a different approach with a pre-heading:
screen shot 2015-05-06 at 10 11 49 am

mlandauer:
You're right. Giving the exact number takes some magic out of it. Go with your original idea of rounding to the hundreds.

mlandauer:
I prefer the version with the preheading. Easier to scan and feels more human.

equivalentideas:
> I prefer the version with the preheading. Easier to scan and feels more human.

Yeah I quite like that too. Happy with that text? 'Find data and code you can use' or maybe 'Find data and code to use'. I think mentioning the "you" in there somehow is nice.

equivalentideas:
I can't get [number_to_rounded()](http://www.rubydoc.info/docs/rails/4.0.0/ActiveSupport/NumberHelper:number_to_rounded) to work for some reason.

In consol I get:

> number_to_rounded(111.2345, precision: 1, significant: true)
> NoMethodError: undefined method `number_to_rounded' for main:Object

mlandauer:
Here's a trick to make that work. The function you're trying to call is a
view helper so is normally only available in views. To be able to access
those in the console you need to do a little bit of a trick (which is also
the same trick you use for calling them in tests) which is to call them
through a `helper` object. So, specifically,

```
helper.number_to_rounded(111.2345, precision: 1, significant: true)
```

equivalentideas:
Thanks @mlandauer , now I'm getting:

> helper.number_to_rounded(111.2345, precision: 1, significant: true)
> NoMethodError: undefined method `number_to_rounded' for #<ActionView::Base:0x000001014ca660>

It's like it just isn't defined for some reason :S.

I pushed the pre-header:

screen shot 2015-05-06 at 10 52 35 am

mlandauer:
Sorry I should have tested it for myself. You'll need to include the correct library. So in the console,
```
> include ActiveSupport::NumberHelper
 => Object
> helper.number_to_rounded(111.2345, precision: 1, significant: true)
 => "100"
```

mlandauer:
@equivalentideas how do you feel about extracting the search box out into a partial and then using that same partial on the main search page?

equivalentideas:
I think that's a good idea. In that case maybe only show the suggested terms if there's no query.

mlandauer:
Not showing the suggested terms when there's a query is an excellent idea!

equivalentideas:
@mlandauer I still can't get the number rounding to work somehow, following your tip I can get it in consol, but when I add it to the view I get 'undefined method `number_to_rounded' for #<#<Class:0x000001016526e0>:0x000001014e3610>' :S

equivalentideas:
I'll extract that partial now.

mlandauer:
Weird. `number_to_rounded` is not defined as a view helper. It's just in `ActiveSupport`. I feel like I'm missing something

equivalentideas:
@mlandauer do you think we need to make some change to the heading text to include users on the search page?

screen shot 2015-05-06 at 11 32 44 am

mlandauer:
@equivalentideas that sounds like a very good idea. Something that could everywhere?

mlandauer:
@equivalentideas can you push a commit to just extract the partial. So I can make the change for the number of scrapers?

equivalentideas:
@mlandauer done with 816d91b

equivalentideas:
> @equivalentideas that sounds like a very good idea. Something that could work everywhere?

Maybe '

Find data and code you can use
Search over 2200 scrapers and users

mlandauer:
I think that wording has lots of ambiguity and it might get a little too longwinded to word it unambiguously. So, maybe we should pull the heading out of the partial and go back to having different headings on those different pages but stick with the common search box and suggested searches?

equivalentideas:
What do you think about?

screen shot 2015-05-06 at 11 53 03 am

adds a little wrap on very small screens, but not a problem I don't think:

screen shot 2015-05-06 at 11 51 13 am

mlandauer:
Unless we change the heading to
> Search over 2200 scrapers and 2000 users

Does that work or is that too long winded?

equivalentideas:
I think too long and we go with your suggestion of going back to the separate headings "Search over 2200 scrapers" and just "Search" on the search page.

> So, maybe we should pull the heading out of the partial and go back to having different headings on those different pages but stick with the common search box and suggested searches?

It might be simpler to do the heading text as an option on the partial, because it's actually the label for the form input, so we'd be losing functionality by pulling it out of the partial because we couldn't make it the label anymore. Does that make sense?

equivalentideas:
I just noticed the search banner form input has the same input as the header search. That's a problem as it's invalid markup and also could confuse assistive tech.

mlandauer:
How about this?

We pull the subheading out of the partial and make the partial just show "search" as the heading but you can optionally override it with your own text that you pass to the partial.

That way the partial stays generic and doesn't get tied into the specific wording and implementation of the search box on the landing page.

Does that make sense?

equivalentideas:
Yep. And would we pass the subheading through the option?

mlandauer:
The subheading would get pulled out of the partial and put just back on the landing page template

equivalentideas:
> The subheading would get pulled out of the partial and put just back on the landing page template

That's gonna require a bit of css rejig to get that to work and it'll be less semantically grouped html, with part of the heading/form label outside of the form block. Feels wrong, maybe not sharing the partial would be simpler? I'm not sure.

mlandauer:
Okay, another solution (which would be my preferred option) as it's the simplest would be to use a separate generic label which just says "Search" and which is only visible to screen readers. Then, the heading (and the sub-heading) can be moved above the form and out of the partial and we can go back to having separate headings on those two pages.

equivalentideas:
> Okay, another solution (which would be my preferred option) as it's the simplest would be to use a separate generic label which just says "Search" and which is only visible to screen readers. Then, the heading (and the sub-heading) can be moved above the form and out of the partial and we can go back to having separate headings on those two pages.

I can live with that. Having a little more complex markup for simpler code. I'll move the 'banner' wrapper stuff and the headings out of the partial, and rename it the search form partial :)

equivalentideas:
just going to rename that partial and do a small refactor

equivalentideas:
I think this is ready to go @mlandauer

Issue #710, opened 2015-05-07
I expect the first url on the api page to be the one I can copy and use


equivalentideas:
When I go to use a scrapers api, I always immediately copy the first link on the ['How to use the api'](https://morph.io/documentation/api?scraper=tupinek%2Fpribor_mesicnik) page before realising that's the example and it doesn't change to match the scraper I'm looking at.

Not sure if this is just an issue for me.

We could make the first url dynamic as well as the one at the bottom of the page you currently use.

mlandauer:
@equivalentideas I think that's excellent feedback. The first url is there to show you what parameters go where. The second url is live and gets filled in automatically.

The first url is really useful when you're developing an app that consumes the api. The second url is really useful to get a handle with a specific example to see how it works.

I think there's value in having both of those there. I think what's wrong and confusing is that they should both be together and the difference between them made clear and explicit.

benrfairless:
Thanks for raising this, and sorry it has sat without a response for so long.

We are tidying up the morph.io issue tracker. This issue was opened in 2015 and has had no activity since, so we are closing it automatically as part of a backlog review to keep the tracker focused on work we can realistically act on. Closing it is not a judgement that the underlying idea was wrong, just that it has gone stale.

If this is still relevant to you, please reopen it or open a fresh issue with up to date detail and we will gladly take another look.

This triage was carried out with AI assistance (Claude Opus 4.8 via Claude Code) and reviewed by the OpenAustralia Foundation team.

PR #711, opened 2015-05-07
[WIP] Notice and documentation that downloads are public

equivalentideas:
We want to give users some notice and explain why we're showing their data download activity.

This PR adds a small notice to the scraper page (next to the download links) and API docs page. It also adds a new section to the documentation about this feature.

I think think notices could be in an into alert style of some kind, gonna try that. Feed back on that and on the text would be much appreciated.

Here's the text file https://github.com/openaustralia/morph/blob/downloader-notice/app/views/documentation/_open_downloads.md

### DIFF
--- a/app/assets/stylesheets/scrapers.css.scss
+++ b/app/assets/stylesheets/scrapers.css.scss
@@ -127,16 +127,23 @@
 }
 
 .scraper-download-block {
-  .btn-group a:last-child {
-    @media (min-width: $screen-md-min) {
-      margin-right: .5em;
+  margin-top: 1em;
+
+  .btn-group {
+    margin-bottom: 1em;
+
+    a:last-child {
+      @media (min-width: $screen-md-min) {
+        margin-right: .5em;
+      }
     }
   }
 }
 
 .download-privacy-notice {
-  margin-top: 1em;
+  margin: 0 0 1em;
   display: inline-block;
+  vertical-align: top;
 
   &.alert {
     padding: 0.25em .5em;
--- a/app/assets/stylesheets/scrapers.css.scss
+++ b/app/assets/stylesheets/scrapers.css.scss
@@ -126,6 +126,14 @@
   }
 }
 
+.scraper-download-block {
+  .btn-group a:last-child {
+    @media (min-width: $screen-md-min) {
+      margin-right: .5em;
+    }
+  }
+}
+
 .download-privacy-notice {
   margin-top: 1em;
   display: inline-block;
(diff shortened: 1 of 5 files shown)


equivalentideas:
I've added a notice style to the notices to make them more prominent and because we use that style for other notices on the site, so we're building up a familiar pattern:

screen shot 2015-05-07 at 5 13 43 pm
screen shot 2015-05-07 at 5 13 55 pm

mlandauer:
I think the alert pattern is good. On the API page I think it's now too prominently placed and the more I think about it, it should go further down the page because the first part of that page is just explaining how to use the api and it just muddles that if it's put where it is now.

Perhaps it should go just above the example?

mlandauer:
I'm having a go at the documentation wording now. It might take me a bit to get into the swing of it.

equivalentideas:
> Perhaps it should go just above the example?

Between the params and example sections? That seems right as even if you're not using the example it's very clear. I always get confused on that page as to what's the bit you use ( #710 ).

equivalentideas:
I'm just committing an improvement to the alert layout css on above the data-table.

mlandauer:
@equivalentideas yes, between the params and example section

mlandauer:
@equivalentideas now that I think of it, based on what you said in #710, maybe that's not the best place. Oh never mind, let's not worry about it and deal with #710 soon and then come back to this

equivalentideas:
> let's not worry about it and deal with #710 soon and then come back to this

That sounds good :+1: .

equivalentideas:
Should I still move it down for now?

equivalentideas:
In making the layout more stable I also just adding a little space about the 'sign in to download' stuff for signed out users:

## Before
screen shot 2015-05-07 at 5 36 40 pm

## After
screen shot 2015-05-07 at 5 36 26 pm

mlandauer:
I don't know. It's not obvious to me where it should go right now. What do you think?
> Should I still move it down for now?

equivalentideas:
It's still quite clear moved down:
screen shot 2015-05-07 at 5 41 01 pm

Some people (particularly small screens) will need to scroll. Hmmm. Maybe we should lean to more obvious for now and not move it.

mlandauer:
Would it be too cheesy to add a lightbulb icon to the alert? http://fortawesome.github.io/Font-Awesome/icon/lightbulb-o/

equivalentideas:
> Would it be too cheesy to add a lightbulb icon to the alert?

I'm not sure if it's the right icon. I think first pass just text.

PR #715, opened 2015-05-08
Refactor email alerts

equivalentideas:
This refactors the alert mailer and process_alerts method to use scrapers instead of runs.

### DIFF
--- a/app/models/scraper.rb
+++ b/app/models/scraper.rb
@@ -38,7 +38,7 @@ class Scraper < ActiveRecord::Base
   extend FriendlyId
   friendly_id :full_name
 
-  delegate :queued?, :running?, to: :last_run, allow_nil: true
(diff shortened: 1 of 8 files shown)


app/models/scraper.rb
```diff
@@ -114,6 +114,22 @@ def latest_successful_run_time
     latest_successful_run.finished_at if latest_successful_run
   end
 
+  def last_run_finished_successfully?
+    if last_run
+      last_run.finished_successfully?
+    else
```
henare: You could make this method much more concise by changing it to just

``` ruby
last_run && last_run.finished_successfully?
```

And same with the next method.

equivalentideas: Do you mean to:

```
def last_run_finished_successfully?
  last_run && last_run.finished_successfully? true : false
end
```

henare: No, I mean what I wrote.

PR #727, opened 2015-05-13
Only show scrapers with data in search by default and toggle with filter buttons.

### LINKED ISSUE #712: Search filter for scrapers with some data
The idea has come up of allowing sorting of scrapers in search results by the amount of data they have. The problem with this is that more data is not necessarily better than little data. What's really important is whether the scraper has any data at all as occasionally people will start writing scrapers but not actually get to the point where it writes anything in the database.

So, it would be good to by default filter scrapers by whether they have any data.

Something like
"Only show scrapers with data"

And a checkbox that is by default on.

equivalentideas:
Help people find functioning scrapers and data more quickly by hiding scrapers without data in search results by default. Provide buttons to toggle the filtering on and off.

These commits change the default search query for scrapers, so it only searches scrapers with data. They add buttons to the top of search results so you can still see scrapers without data if you need to.

This does mean there's an extra search query taking place in the search controller, there may be a way to refactor that, but I couldn't work it out.

closes #712

screen shot 2015-05-12 at 4 54 01 pm
screen shot 2015-05-13 at 10 30 04 am
screen shot 2015-05-13 at 10 36 42 am
screen shot 2015-05-13 at 10 36 55 am

I wasn't sure about having the buttons to the right on wide screens at first, but I've come around to it. It makes these options something you reach for if you need, rather than a central part of finding scrapers and data. I think this is right, as people just what to find stuff, not explore all the search options.


equivalentideas:
Just noticed that 'no results for scrapers with data' message is a bit confusing:

> Sorry, we couldn’t find any scrapers *with data* relevant to your search term **“bugzappers”**.

sounds like that data wasn't relevant, not the scraper.

equivalentideas:
> Sorry, we couldn't find any scrapers relevant to your search term **"bugzappers"** that have data.

> Sorry, we couldn't find any scrapers relevant to your search term **"bugzappers"** with data. Try searching all scrapers.

> Sorry, we couldn't find any scrapers with data, that are relevant to your search term **"bugzappers"**.

I think this last one is a bit clearer with the comma and 'that' to break up these concepts. I'll commit that and we can keep iterating if need be.

mlandauer:
Looking at the screenshots it surprises me to see the "scrapers with data/all scrapers" button outside the scrapers tab. Isn't that breaking the simple physical metaphor that the tab provides?

In the tab metaphor we're hiding and showing whats below the tab navigation because they are "separate" sheets of paper.

In the current implementation you're showing and hiding the buttons when they're outside that virtual "sheet" as if they were on that sheet.

To keep the metaphor consistent wouldn't be it more consistent to have the "scrapers with data/all scrapers" buttons below the tab navigation rather than beside it? It could still be off on the right hand side of course.

spec/models/scraper_spec.rb
```diff
@@ -183,6 +183,22 @@
...
 
+    context "scraper has no data" do
+      describe "#has_data?" do
+        it{expect(scraper.has_data?).to eq false}
+      end
+    end
+
+    context "scraper has data" do
+      before :each do
+        allow(scraper).to receive(:sqlite_db_size).and_return(256000)
```
mlandauer: As far as I'm aware it's perfectly possible to have an sqlite database with a non-zero size that has no data in it. So, I don't think this is the best way to measure this

app/models/scraper.rb
```diff
@@ -44,10 +44,15 @@ def search_data
     {
       full_name: full_name,
       description: description,
-      scraped_domain_names: scraped_domain_names
+      scraped_domain_names: scraped_domain_names,
+      has_data?: has_data?
```
mlandauer, line 48: How would this get reindexed if the amount of data the scraper has changes?

app/controllers/search_controller.rb
```diff
@@ -1,8 +1,16 @@
...
     @type = params[:type]
+    @show = params[:show]
+
+    if @show == "all" && @type != "users"
+      @scrapers = Scraper.search @q, fields: [{full_name: :word_middle}, :description, {scraped_domain_names: :word_end}], highlight: true, page: params[:page], per_page: 10
+      @scrapers_total_count = @scrapers.total_count
+    else
+      @scrapers = Scraper.search @q, where: {has_data?: true}, fields: [{full_name: :word_middle}, :description, {scraped_domain_names: :word_end}], highlight: true, page: params[:page], per_page: 10
+      @scrapers_total_count = (Scraper.search @q, fields: [{full_name: :word_middle}, :description, {scraped_domain_names: :word_end}]).count
+    end
```
mlandauer, line 19: This is too complicated

equivalentideas:
> To keep the metaphor consistent wouldn't be it more consistent to have the "scrapers with data/all scrapers" buttons below the tab navigation rather than beside it? It could still be off on the right hand side of course.

I agree this makes sense. As above, I was trying to reduce the prominence of the filters by moving them out of the lefthand line of reading. And then moving it up so you don't get a weird s shape in the white space to the left. This is simple, but it does break the page metaphor.

The way other search interfaces seem to deal with this is by adding a sidebar for facets/filtering.

screen shot 2015-05-13 at 11 33 39 am

Thinking about this more, maybe the simpler approach would be to move away from the tab pattern. Which we don't use anywhere else in the app.

mlandauer:
Adding a sidebar makes sense to me. I don't see what the issue with that is.

app/views/search/search.html.haml
```diff
@@ -4,36 +4,46 @@
...
+          - if @type != "users" && !@scrapers_total_count.zero?
+            .list-filters.btn-group
+              = link_to "Only scrapers with data", search_path(q: @q), class: "btn btn-default #{('active' unless @show == 'all')}"
+              = link_to "All scrapers", search_path(q: @q, show: "all"), class: "btn btn-default #{('active' if @show == 'all')}"
           - @scrapers.with_details.each do |scraper, details|
             = render scraper, highlight: details[:highlight]
-        = paginate @scrapers
+          = paginate @scrapers
+
+        - if @type == "users" && @owners.empty? && !@scrapers_total_count.zero?
```
mlandauer: This whole block could do with some refactoring

equivalentideas:
I'm going to work on the layout alternative ideas then work those the comments on the commits.

equivalentideas:
A few alternative layouts we could use:

1. A sidebar for all the search options (including types) on the left, like Github, which removes the tab styles: 


2. Settings for this search over to the right, under the tab, similar to duck duck go.



3. Settings small, underneath the result types like google:
screen shot 2015-05-14 at 10 34 15 am

I think the Option 2 is the simplest iteration from where we are, and keeps the settings as a clear secondary interaction. Option 1 could be good, but it's a bigger change to include this one thing. Option 3 could be good if we had more settings to toggle, but that's not a problem to solve yet.

I'm going to see how Option 2 works implemented.

mlandauer:
Option 2 is definitely the best right now

equivalentideas:
Nice, here's what that could look like:

screen shot 2015-05-14 at 11 14 53 am

mlandauer:
Looks good. "Search tools" is fairly meaningless. Either there shouldn't be a title or it should be something more meaningful and specific like "Filter"

equivalentideas:
I think the a title is helpful for speeding up understanding in this layout. I just took 'search tools' as a default from google, but 'Filter' or 'Filter search' is better as it describes the action :+1: .

mlandauer:
Also, seeing the buttons in place I think their text can be shortened to "with data" and "all" which actually makes them quicker to scan. "Only" isn't really necessary.

If we really want to keep scrapers somewhere around there then it should be probably be in the title which could be "Filter scrapers"

equivalentideas:
^^^ That was a bit epic. Ended up the most maintainable way to do that was using the bootstrap grid classes, but the meant it had to be applied to everything in the .search-results wrapper. Now adding the heading and making changes to the button text.

That view is in dire need of refactoring.

equivalentideas:
screen shot 2015-05-14 at 12 41 47 pm

I like the shorter button text, but there's something no quite right about 'Filter scrapers: All' as that's not really a filter. Maybe "Off'. Would be better. Gonna look for some examples quitckly.

equivalentideas:
Maybe a toggle button would be better, but feels like a next pass thing:

equivalentideas:
I hid the heading on small screens as it sits under the scraper heading and I think it's much simpler without it.

screen shot 2015-05-14 at 1 05 21 pm

spec/models/scraper_spec.rb
equivalentideas: You're right, and I missed that there's a method on scrapers `sqlite_total_rows` :hushed: . That might be more useful. In my local version it's definitely missing a scraper with data. Thanks!

app/models/scraper.rb
equivalentideas, line 48: Searchkick reindexes this by default, like the scraped domains, and I've tested it reindexes when a scraper that has no data gets run and then has rows.

It wasn't getting updated when a scraper db get's cleared, so I've added reindex to that action (https://github.com/openaustralia/morph/commit/3286fcb72407513622b24878fed45bbcd1acfe1f) which does the trip.

There's still the situation you describe where a run changes the number of rows to 0. I can't actually work out a way to test if this is happening. That does seem like a real corner case to me, but what do you think?

mlandauer:
Filter state isn't maintained after switching between tabs. To reproduce this:

* Go to http://127.0.0.1:3000/search?q=foo&show=all. This shows all scrapers matching "foo".
* Click on the "users" tab
* Click on the "scrapers" tab

I would expect to be back on the same page but instead I am on the page http://127.0.0.1:3000/search?q=foo which is different

app/views/search/search.html.haml
```diff
@@ -2,38 +2,29 @@
...
+            %aside.col-xs-12.col-md-3
+              = render partial: "scraper_search_filter_controls"
+            - # If there are scrapers matching filters
+            - if !@scrapers.total_count.zero?
+              .col-xs-12.col-md-9
+                = render partial: "search_results_list", locals: {results: @scrapers}
+            - else
+              = render partial: "no_results_of_type_notice"
+          - else
+            = render partial: "no_results_of_type_notice"
```
mlandauer: By splitting this out into a partial you've separated two bits that deal with similar things - which is what to show when there are no results. These are now in the partial "no_results_of_type_notice" and in the code below. General guiding principle: Similar things should be close together.

equivalentideas:
> I would expect to be back on the same page but instead I am on the page http://127.0.0.1:3000/search?q=foo which is different

I think the same applies if you do a new search as well. At the moment it resets to only scrapers with data.

equivalentideas:
I've updated it to maintain the 'show' param for filters when you navigate between types and submit a new search.

Thanks for that feedback @mlandauer it definitely makes the process smoother.

I think this is ready to merge if there's nothing else :rocket:

mlandauer:
@equivalentideas it's for me to say when a PR is ready to merge! :-)

equivalentideas:
:+1:

app/controllers/search_controller.rb
```diff
@@ -1,8 +1,21 @@
...
-    @owners = Owner.search @q, highlight: {fields: [:nickname, :name, :company, :blog]}, page: params[:page], per_page: 10
     @type = params[:type]
+    @show = params[:show]
+
+    @owners = Owner.search @q, default_owner_search_params
+
+    all_scrapers = Scraper.search @q, default_scraper_search_params
+    @unfiltered_scrapers_count = all_scrapers.total_count
+
+    if @show == "all" && @type != "users"
```
mlandauer: You shouldn't actually need to test what the @type is here, right? Okay, it makes the code very slightly less efficient but it also makes it slightly more complicated

equivalentideas: You're right, it's not needed. I don't think it really makes it more efficient either. At the moment if you're on the users results with without @show, it makes a second unused search query for scrapers.

So I think for starters we could remove that `@type != "users"`. If we want to stop it making that unnecessary query we could wrap that whole bit in `if @type != "users". Maybe that's over optimising?

mlandauer: Clarity pretty much always trumps efficiency, so yes, please remove `@type != "users"`

equivalentideas: Removed in 2af7127

PR #736, opened 2015-05-18
Summarise scraper for scraper show page meta description

### LINKED ISSUE #663: meta info for each scraper page
Make the meta info for a scraper page be a simple summary of that scraper and possibly the domains that it scrapes. This would be much better than what it currently does which is just the default site-wide meta info about morph.io

equivalentideas:
To improve the indexing and display of scrapers in search engines, these commits change the description of a scraper page to contain the description the owner has provided.

If they haven't added a description it falls back to a generic description "A scraper to collect structured data from the web."

If they haven't added a description, *but the scraper has a list of scraped domains*, then it adds that as well to make "A scraper to collect structured data from foo.com, bar.com and fiddle.com."

This uses the meta-tags gem and removes the hardcoded description from the application layout. This means that pages without the description set wont have any description anymore. I've added the old description back to the signed_out landing page. Aside from that, I don't think it's a regression to remove the current description from other pages, because it isn't actually a description of the page and so isn't useful to search bots or humans (if they can even see it somewhere).

fixes #663

### DIFF
--- a/app/views/static/_signed_out_index.html.haml
+++ b/app/views/static/_signed_out_index.html.haml
@@ -1,3 +1,5 @@
+- set_meta_tags description: 'Get structured data out of the web. Code collaboration through GitHub. Run your scrapers in the cloud.'
+
 .banner-wrapper
   .banner.container#banner
     %h2 Take the hassle out of web&nbsp;scraping
--- a/app/views/static/_signed_out_index.html.haml
+++ b/app/views/static/_signed_out_index.html.haml
@@ -1,5 +1,3 @@
-- set_meta_tags description: 'Get structured data out of the web. Code collaboration through GitHub. Run your scrapers in the cloud.'
-
 .banner-wrapper
   .banner.container#banner
     %h2 Take the hassle out of web&nbsp;scraping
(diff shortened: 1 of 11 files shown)


equivalentideas:
@mlandauer I've added the helper (with tests) to show the list of domains in the description if it can. I've refactored it as well.

Could you please test this works with domains locally? I can only test the has 'description' and 'no descriptions' cases, but not the 'no description, has domains list' situation without the domains data locally.

Thanks!

app/views/static/_signed_out_index.html.haml
```diff
@@ -1,3 +1,5 @@
+- set_meta_tags description: 'Get structured data out of the web. Code collaboration through GitHub. Run your scrapers in the cloud.'
```
mlandauer: It doesn't make sense for a url to have different meta tags in different circumstances. So, this should be set on `app/views/static/index.html.haml` so that it applies whether you're logged in or not.

mlandauer:
Just the one bit of feedback to address. Other than that I think it's good to merge :-)

equivalentideas:
Excellent, I'll jump on that after https://github.com/openaustralia/morph/issues/743

app/views/static/_signed_out_index.html.haml
equivalentideas: Thanks @mlandauer I've implemented your suggestions with https://github.com/openaustralia/morph/commit/b802a96a9477f2598f5b444032ac7215597a60e3

Issue #737, opened 2015-05-18
Scroll bar visible next to filter buttons


henare:
Firefox on Linux, of course:

![selection_001](https://cloud.githubusercontent.com/assets/48945/7677260/cf688dca-fd4a-11e4-8218-a08c80e6bb6a.png)

equivalentideas:
Thanks for catching this @henare 

This is related to using ['overflow' to property to hide the heading on small screens](https://github.com/openaustralia/morph/blob/master/app/assets/stylesheets/search.css.scss#L35).

I can't reproduce this on my machine, could you please try changing the [overflow value for wide screens](https://github.com/openaustralia/morph/blob/master/app/assets/stylesheets/search.css.scss#L40) to visible or hidden and see what the impact is?

Either of those values should be fine for the layout, but I'm not sure if they will fix this bug.

henare:
>I can't reproduce this on my machine, could you please try changing the overflow value for wide screens to visible or hidden and see what the impact is?

Both those work.

equivalentideas:
Great thanks for that :+1:

PR #741, opened 2015-05-19
Fix empty scrollbar bug effecting Linux users

### LINKED ISSUE #622: Data table has empty scroll bars
https://morph.io/openaustralia/nsw_agencies_from_directorynswgovau

selection_001

equivalentideas:
Based on feedback from @henare this should remove scroll bars
that would appear around the data table even if it didn't need
scrolling. This fix doesn't effect Chrome or Firefox on OSX, which
aren't experiencing the bug.

Fixes #622


henare:
Confirmed fixed. #737 still an issue.

PR #753, opened 2015-05-22
Hide owner option from new scraper form if user only has one option

equivalentideas:
This removes the 'owner' heading and select menu from the form if the user doesn't have a choice, so they can focus on the inputs they can interact with.

It also removes the '/' prepended to the input if the owner block isn't there, because it is confusing to have it present if it isn't dividing anything.

closes #753 

screen shot 2015-05-22 at 3 58 46 pm

### DIFF
--- a/app/views/scrapers/new.html.haml
+++ b/app/views/scrapers/new.html.haml
@@ -13,13 +13,14 @@
           = f.input :owner_id, collection: current_user.all_owners.map{|o| [o.nickname, o.id, {"data-content" => "#{owner_image(o, 20, false)} #{o.nickname}"}]}, include_blank: false
       .col-sm-6
         = f.input :name do
+          - name_input_field = f.input_field :name, class: "form-control", placeholder:"city_of_sydney_development_applications"
           - if current_user.all_owners.count == 1
-            = f.input_field :name, class: "form-control", placeholder:"city_of_sydney_development_applications"
+            = name_input_field
           - else
             -# Using long-winded form here because simple_form doesn't directly support prepends
             .input-group
               %span.input-group-addon /
-              = f.input_field :name, class: "form-control", placeholder:"city_of_sydney_development_applications"
+              = name_input_field
           %span.help-block Name to give the scraper here and on GitHub
 
     .row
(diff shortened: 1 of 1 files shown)


mlandauer:
@equivalentideas have you actually tested that this works (i.e. creates a scraper correctly) in the situation where the owner select is removed from this form? Looking at the code it looks to me like you're just removing the field from the view and the controller has no special handling for that.

equivalentideas:
You're right :S doesn't work. I should have tested for that! I'll fix that up.

equivalentideas:
I've now testing this works property, sorry about before. I added the owner input back in and made it a hidden input if their is only 1 potentially owner to choose from.

app/views/scrapers/new.html.haml
```diff
@@ -6,15 +6,20 @@
...
-        -# Using long-winded form here because simple_form doesn't directly support prepends
-        = f.input :name do
-          .input-group
-            %span.input-group-addon /
-            = f.input_field :name, class: "form-control", placeholder:"city_of_sydney_development_applications"
-          %span.help-block Name to give the scraper here and on GitHub
+      - if current_user.all_owners.count == 1
+        = f.input :owner_id, as: "hidden", input_html: { value: current_user.id }
+        .col-sm-6
+          = f.input :name, placeholder: "city_of_sydney_development_applications", hint: "Name to give scraper here and on Github"
```
mlandauer: There's now duplication of the placeholder and hint text between those two variants of it. Since simple_form doesn't support prepends one of the variants isn't using simple_form. If you write this variant in its long form not using simple_form, you'll probably find that you can simplify this a lot and at the very least remove the duplication of placeholder and hint.

Whenever you see that kind of duplication (and even more so when you create it!) you should work hard to get rid of it.

equivalentideas: Thanks for this, I'll clean that up.

equivalentideas:
@mlandauer I've removed that duplication here by using a temporary variable. I was trying to work out some way to inline it again, but couldn't work it out. I do think this is a bit more readable though with the separate inputs grouped together.

PR #764, opened 2015-05-25
[For review] Basic running scrapers page and remove 'recently active scrapers' list from landing page

### LINKED ISSUE #757: Make new page to view running scrapers
This is because we want to remove this list of recently active scrapers from the landing page. It's still useful for admins to see this, so we're going to put it on it's own page. Initially we can link to it from the admin dropdown in the menu.

equivalentideas:
This adds a new page `/scrapers/running` that you can access through the admin dropdown menu. The page shows a list of scrapers currently running and has a heading. Nothing fancy.

@mlandauer This is what's described in issue #757 , but it's different functionality than you currently get from the 'Recently active scrapers' list on the landing page. Does this list of just the currently running scrapers satisfy your needs?

I haven't added a test for the `Scraper.currently_running` class method, but I also see that we don't have testings for `.running?`. I read that it's better to test the direct behaviour that the association, so should I add a tests for the [`.running?`](https://github.com/openaustralia/morph/blob/master/app/models/run.rb#L63-L65) method on Run?

Also this instantiates Scraper.all to get this list so it might be a bit slow on production. I couldn't work out a more performant way to do this sorry.

closes #757 

screen shot 2015-05-25 at 7 12 24 pm
screen shot 2015-05-25 at 7 12 17 pm

### DIFF
--- a/app/controllers/scrapers_controller.rb
+++ b/app/controllers/scrapers_controller.rb
@@ -228,6 +228,7 @@ class ScrapersController < ApplicationController
   end
 
   def running
+    @scrapers = Scraper.currently_running
   end
 
   private
--- a/app/controllers/scrapers_controller.rb
+++ b/app/controllers/scrapers_controller.rb
@@ -227,6 +227,9 @@ class ScrapersController < ApplicationController
     authorize! :watchers, @scraper
   end
 
+  def running
+  end
+
   private
 
   def render_error(message)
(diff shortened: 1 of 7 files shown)


equivalentideas:
I'm going to remove the scrapers list on the landing page (#767) on this PR because that was the reason to do #757 in the first place. It makes sense to me that they are merged together.

Issue #767, opened 2015-05-26
Remove 'Recently active scrapers' from landing page


equivalentideas:
The 'recently active scrapers' list on the landing page is not useful to people using morph.io. Just because a scraper is running or has run recently, that doesn't make it interesting or relevant to someone.

It is there to serve the administrators. Once #757 is fixed (so admins are still served properly) we can remove it from the landing page.

Issue #751, opened 2015-05-22
Style the scale metrics


equivalentideas:
Here's how this block currently appears:

![screen shot 2015-05-22 at 11 12 29 am](https://cloud.githubusercontent.com/assets/1239550/7762170/d33cefbe-0073-11e5-8016-a56e8d005738.png)
![screen shot 2015-05-22 at 11 13 34 am](https://cloud.githubusercontent.com/assets/1239550/7762171/d341bec2-0073-11e5-983d-f44e349717e4.png)
![screen shot 2015-05-22 at 11 14 33 am](https://cloud.githubusercontent.com/assets/1239550/7762172/d34b17f6-0073-11e5-9b55-3072ecaca1ff.png)

equivalentideas:
These stats (see #119 ) are meant to be:

> demonstrating the scale of activity on morph.io to validate and reinforce that morph is awesome and they should give it a go or explore it.

I think that these currently appear too loud and in-your-face. Morph's character should be clear, simple, focuses on helping you do what you want, with a bit of fun thrown in. I think these stats should be salient enough that you see them and the numbers have their impact on you, but clearly supporting information.

I think turning them into a ribbon above the footer, maybe with colour to pop them out, and reducing their size quite a bit could be a good way to go. I'll sketch out some ideas around that.

mlandauer:
@equivalentideas remember why those stats are there

equivalentideas:
@mlandauer totally, I think we still want people to say 'wow that's awesome' but I think we can make them obvious without making them so big.

mlandauer:
It's at a size that's in keeping with other elements on the landing page when you're not signed in. It seems strange to me to want to significantly reduce the size.

equivalentideas:
I'll do a quick go at what I'm thinking, very happy to ditch it if it doesn't work ;)

equivalentideas:
This is the kind of concept I was trying to describe:
![screen shot 2015-05-22 at 11 42 02 am](https://cloud.githubusercontent.com/assets/1239550/7762420/a648c056-0077-11e5-9ff3-303490ee841e.png)

equivalentideas:
Type size is pretty similar actually.

mlandauer:
@equivalentideas so you've mostly reduced the whitespace around the text, here? I'm surprised, you being a fan of whitespace and all. ;-)

equivalentideas:
[...]

PR #768, opened 2015-05-26
[WIP] New users list

### LINKED ISSUE #758: Remove organisations list from the landing page

### LINKED ISSUE #759: Make the view users list on the landing page more focused on the pictures
"When I log in, show me a list of new users who've joined morph.io, so I can connect with them and discover interesting work, and feel like morph is growing."

Make the pictures big and only include people with pictures.

We're thinking we'll only show the users with actual avatars, not the default gravatar shapes, because it's very difficult to get any information or be engaged by avatars people haven't set themselves.

equivalentideas:
This reworks the new users list to put more focus on the people. It makes the images bigger, and removes the names. You can still get the user's nickname with a tooltip on hover.

It removes the organisation list, because this list is very static and not very useful to users.

The users list has it's own row, so the the scrapers list (which we're soon to remove, see #767 ) slots in underneath.

**TODO: Don't show users who only have the generated identicon**, like screen shot 2015-05-26 at 3 31 01 pm

The purpose of this change is undermined a bit if there are lots of the identicon avatars. It makes the users seem like robots even! I can't work out how to filter them out unfortunately. Any ideas @mlandauer ?

closes #759 
closes #758 

## Before
screen shot 2015-05-26 at 3 32 09 pm

## After
screen shot 2015-05-26 at 3 32 49 pm

### DIFF
--- a/app/assets/stylesheets/_landing_page.scss
+++ b/app/assets/stylesheets/_landing_page.scss
@@ -263,6 +263,10 @@
   }
 }
 
+.new-users-block {
+  margin: 24px 0;
+}
+
 .banner-metrics {
   padding-bottom: 24px;
   text-align: center;
--- a/app/assets/stylesheets/users.css.scss
+++ b/app/assets/stylesheets/users.css.scss
@@ -41,3 +41,9 @@
     font-size: $font-size-large;
   }
 }
+
+.owner-list-simple {
+  .owner-item {
+    margin-bottom: $padding-large-vertical;
+  }
+}
--- a/app/views/static/_signed_in_index.html.haml
+++ b/app/views/static/_signed_in_index.html.haml
@@ -15,7 +15,7 @@
   = render "broken_scrapers"
 
   .row
-    .col-xs-12
+    .new-users-block.col-xs-12
       %h2
         %small.pre-heading Recent new users
         Welcome to morph.io
--- a/app/views/static/_signed_in_index.html.haml
+++ b/app/views/static/_signed_in_index.html.haml
@@ -21,7 +21,7 @@
         - User.all.order(created_at: :desc).limit(10).each do |user|
           %li
             = link_to user do
-              = owner_image(user, 20)
+              = owner_image(user, 60)
       %p
         = link_to users_path do
           See more of the #{pluralize(User.count, "user")}&hellip;
--- a/app/views/static/_signed_in_index.html.haml
+++ b/app/views/static/_signed_in_index.html.haml
@@ -17,7 +17,7 @@
   .row
     .col-xs-12
       %h2 Recent New Users
-      %ul.list-unstyled
+      %ul.list-unstyled.list-inline
         - User.all.order(created_at: :desc).limit(10).each do |user|
           %li
             = link_to user do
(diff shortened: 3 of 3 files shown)


mlandauer:
It's not obvious to me either a simple way you can tell whether a user has an identicon rather than an avatar that they've uploaded themselves. The best I can come up with so far is either to check that the image is a small png (file size) or that the image matches `https://github.com/identicons/<name>.png` where `name` is their github user name.

I hope that's not the only way to do it because that would be nasty and too much work, frankly. We'd have to add another field to the owners table and create a background job that checks and updates those values. Hideous!

@equivalentideas you said (in chat, I think) that you'd been in touch with github to ask them. Where did you ask them? I didn't see anything.

equivalentideas:
> @equivalentideas you said (in chat, I think) that you'd been in touch with github to ask them. Where did you ask them? I didn't see anything.

I asked them on twitter which usually works well for me, but they haven't replied, so I've now put in a support ticket through their form.

Hmmm. Comparing the images, there's no consistent size (file size) difference between the identicons and the images, sometimes the identicons are bigger even. But—the identicons are all 420x420 resolution. That seems to be something consistent we could test for. It still seems very nasty.

I wonder if there's a proxy indicator we could use—like users with photos are more likely to have full names or something. Not a great solution, but I'll have a quick look to see if there's anything that could work.

equivalentideas:
So with a look through the admin, the users with images consistently have full names, and the identicons consistently don't. There are exceptions but it is more than 80:20 I'd say.

mlandauer:
I'd be happy showing avatars for people with their full names filled out. I think that would work "well enough". Could you implement that?

equivalentideas:
Yep I'll go for it :+1: Seems like a decent stop-gap for now.

mlandauer:
I'll deploy to production after this PR is merged

equivalentideas:
Now that we've moved this page to a left aligned layout feel, rather than the 50/50 we had before, it's feeling a bit off balanced with the metrics being centred I think. The full width scrapers in need of help is also a bit odd when there's only one. I feel that has more to do with #751 , not a blocker on this.

screen shot 2015-05-27 at 3 15 38 pm

mlandauer:
The first thing that comes to mind for me looking at the new layout deployed in production is that I want to see the new users across the full width of the screen. only seeing 8 seems so measly! :-)

equivalentideas:
One consideration for having a smaller number was that the list will cycle though a bit more quickly and it feels like there are lots of new users. But maybe we should just get some more users ;)

Here's with 15:
screen shot 2015-05-27 at 3 45 48 pm
screen shot 2015-05-27 at 3 45 41 pm
screen shot 2015-05-27 at 3 45 34 pm
screen shot 2015-05-27 at 3 45 28 pm

10:
screen shot 2015-05-27 at 3 37 23 pm

12 stretched across:
screen shot 2015-05-27 at 3 40 29 pm

12 bigger:
screen shot 2015-05-27 at 3 41 05 pm

I actually prefer a lower number like 10 that doesn't go all the way across cause it's less overwhelming and I feel like a engage more with the individuals.

equivalentideas:
@mlandauer do you still feel like this after living with it for a while?

> The first thing that comes to mind for me looking at the new layout deployed in production is that I want to see the new users across the full width of the screen. only seeing 8 seems so measly! :-)

I actually quite like the concise list of 8 as it feels like a balanced little unit in the layout. From the perceptive of someone dropping in once in a while, in could be nicer to see a longer list though. I could go either way.

PR #796, opened 2015-06-17
Add Ruby quick start instructions.

emikulic:


mlandauer:
This PR overwrites the contents of another bit of help which I think we still need and fulfils a slightly different purpose. Perhaps this new bit of documentation can go somewhere else? Like maybe a ruby/quickstart rather than being the main ruby documentation?

emikulic:
I've restored the docs I previously merged away. Does this look ok?

mlandauer:
Looks good. Thanks @emikulic!

PR #834, opened 2015-07-07
Rename plan IDs

henare:
@mlandauer said the plan stuff needs refactoring so I'm not sure we want to merge this without doing a refactor.


mlandauer:
Might as well merge this and then refactor

PR #878, opened 2015-07-21
Create default watches on sign up

### LINKED ISSUE #275: By default watch scrapers for you and your organisations

henare:
Fixes #275.

Opening for some feedback from @mlandauer since it feels a little hokey and I think the wording needs some work.


app/views/users/_sign_in_message.html.haml
```diff
@@ -0,0 +1,6 @@
+- if @sign_up
+  Welcome to morph.io! You're
+  = link_to "now watching", watching_user_path(@user)
```
mlandauer: I don't think there's any need to let people know about this at this stage. They've just logged into morph for the first time. I think it's unlikely that this message would make much sense at this stage.

Maybe another way users could be informed is that the very first time they receive an email alert there's an introductory paragraph that explains why they are receiving that email and what they can do to change the settings.

app/views/users/_sign_in_message.html.haml
```diff
@@ -0,0 +1,6 @@
+- if @sign_up
```
mlandauer: Generally good practise to avoid the use of member variables in partials. Better to refactor to `sign_up` with `signup: @sign_up` used in the `render` in the controller.

mlandauer:
This PR doesn't cover the situation where someone gets added to an organisation. I think if that happens they should start getting alerts for that organisation as well.

app/views/users/_sign_in_message.html.haml
henare: Thanks, that makes a lot of sense.

app/views/users/_sign_in_message.html.haml
henare: Very good call.

henare:
Thanks for all the feedback @mlandauer.

henare:
>Maybe another way users could be informed is that the very first time they receive an email alert there's an introductory paragraph that explains why they are receiving that email and what they can do to change the settings.

Before we deploy this we'd have to work out who'd already been sent an alert and not show them this message.

henare:
This is ready for review again.

app/views/alert_mailer/alert_email.html.haml
```diff
@@ -192,6 +192,22 @@
         }
 
   %body(bgcolor="#f6f6f6")
+    - unless @user.alerted_at
```
mlandauer: Extract that as simple method? Something like `User#never_alerted?` then you would end up with something that reads better ... `if @user.never_alerted?`

```
def never_alerted?
  alerted_at.nil?
end
```

mlandauer:
I made a tiny suggestions. I'm bikeshedding really. @henare please merge this when you're happy.

app/views/alert_mailer/alert_email.html.haml
henare: Thanks, that makes sense.

henare:
>Before we deploy this we'd have to work out who'd already been sent an alert and not show them this message.

I've spent a lot of today trying to work out some fancy way to do this. But instead of this approach we can just change the message to not say "this is your *first* alert" and it solves the same problem.

We also need a way to switch on alerts for existing users in a nice way. After playing with a few complex ideas I've realised the simplest and best is to just enable the default watching (i.e. you and all your orgs) for all people that are not currently watching anything.

henare:
OK I think this is ready to go. Once deployed we can run the rake task added in 78b88630db4f28a687407956c43ad50631163ffe and then revert that commit.

I'm not merging and deploying this today because I had a really bad day of coding and made lots of silly mistakes. I either need this reviewed by someone else or by me another day :)

mlandauer:
LGTM

henare:
Deployed.

Ran the rake task and it looks to have created default watches for 2,583 morph.io users. If they have erroring scrapers they should get their first alert email at 11:30 Sydney time tomorrow.

Reverted the commit to remove the rake task.

Issue #886, opened 2015-08-07
Visually distinguish "Clear database" and "Delete scraper"


tmtmtmtm:
![screen shot 2015-08-07 at 09 38 12](https://cloud.githubusercontent.com/assets/57483/9131996/29c0dbf6-3ce8-11e5-9db3-7f05a747bd14.png)

Both of these look the same, are in essentially the same place, and have largely identical "are you sure" popups, but the effects of one are rather more severe than the other.

At least twice now I've managed to completely delete a scraper rather than just emptying it because my brain has no short-cuts to differentiating between them and I wasn't paying sufficient attention. Physically separating them more (even to the left and right of the screen), or even making them each a different colour, would make a big difference for very little effort.

henare:
:+1: :+1: :+1:

I came across this exact same issue a couple of days ago!

mlandauer:
This is what it looks like now 
![image](https://cloud.githubusercontent.com/assets/2221/16508481/a8dae418-3f76-11e6-8063-ebb22498d4d2.png)

Based on that I think we can probably close this now. @tmtmtmtm if you disagree please reopen the ticket.

PR #896, opened 2015-08-28
Distinguish and add more explanation to dangerous settings

equivalentideas:
Why:

@tmtmtmtm summarises the issue well:

> Both of these [the **clear database** and **delete scraper** options] look the same, are in essentially the same place, and have largely identical "are you sure" popups, but the effects of one are rather more severe than the other.
>
> At least twice now I've managed to completely delete a scraper rather than just emptying it because my brain has no short-cuts to differentiating between them and I wasn't paying sufficient attention.

Basically these's so little distinction between the two options that they’re easy to confuse. I also think that aside from the characteristically fun “Here be dragons”, the button text and alert is quite cold.

Making these options more distinguishable helps people avoid deleting their scraper accidentally, and making the options more understandable helps people use them more effectively. 

This PR:

* Move's the important 'these can't be undone' message above the buttons
* Adds more text to explain and distinguish the options
* Uses the text-left buttons-right pattern—familiar from Github—to slow the user down and get the to consider more.
* Makes the text friendlier, more conversational and human overall.

While on #886 it was suggested we just change the buttons colour or layout, when I looked at it, I really felt more text, explanation and a human touch was needed for starts. I think this also distinguishes the options.

## Before:

screen shot 2015-08-28 at 10 44 12 am
screen shot 2015-08-28 at 10 44 23 am

## After:

screen shot 2015-08-28 at 10 43 55 am
screen shot 2015-08-28 at 10 44 49 am

Any thoughts feedback @tmtmtmtm ?

### DIFF
--- a/app/assets/stylesheets/scrapers.css.scss
+++ b/app/assets/stylesheets/scrapers.css.scss
@@ -417,40 +417,32 @@ a i.fa {
 }
 
 .danger-settings-options-list {
-  .list-group-item {
-    padding-bottom: 1em;
-
+  .danger-settings-action-body {
     @media (min-width: $screen-xs-min) {
-      @include clearfix;
+      display: flex;
+      align-items: center;
+      justify-content: space-between;
+      flex-wrap: nowrap;
     }
   }
 
   h4 {
-    @media (min-width: 35em) {
+    @media (min-width: $screen-xs-min) {
       margin-bottom: 0;
     }
   }
 
   p {
     @media (min-width: $screen-xs-min) {
-       margin-bottom: 0;
-       float: left;
-       max-width: 20em;
-    }
-
-    @media (min-width: 35em) {
-       margin-top: $panel-body-padding;
-       max-width: none;
+      flex-shrink: 100;
+      margin-top: $panel-body-padding;
+      padding-right: 2em;
     }
   }
 
   .btn {
     text-shadow: 1px 1px 0 rgba(0, 0, 0, .2);
     letter-spacing: 1px;
-
-    @media (min-width: $screen-xs-min) {
-       float: right;
-    }
 
     i {
       padding-left: .25em;
--- a/app/assets/stylesheets/scrapers.css.scss
+++ b/app/assets/stylesheets/scrapers.css.scss
@@ -451,5 +451,10 @@ a i.fa {
     @media (min-width: $screen-xs-min) {
        float: right;
     }
+
+    i {
+      padding-left: .25em;
+      vertical-align: text-top;
+    }
   }
 }
--- a/app/assets/stylesheets/scrapers.css.scss
+++ b/app/assets/stylesheets/scrapers.css.scss
@@ -433,6 +433,7 @@ a i.fa {
 
   .btn {
     text-shadow: 1px 1px 0 rgba(0, 0, 0, .2);
+    letter-spacing: 1px;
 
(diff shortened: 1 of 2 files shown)


equivalentideas:
Assigning to @henare for review next week.

tmtmtmtm:
These will almost certainly still have the same problems for me. Although they're better explained etc., and thus presumably better for new users, I already _know_ exactly what both do, and still occasionally hit the wrong one — because I'm *not* stopping to carefully read things every time I visit the page. Trying to make users who think they already know what they're doing read the text isn't really the problem. You'll get much better results by giving more points of obvious differentiation — i.e. visually distinguishing by something other than text.

henare:
@equivalentideas I agree with @tmtmtmtm, assigning back to you to see if you can address those points.

equivalentideas:
Sorry it's taken me so long to get back around to this! :rabbit:

> These will almost certainly still have the same problems for me. Although they're better explained etc., and thus presumably better for new users, I already know exactly what both do, and still occasionally hit the wrong one — because I'm not stopping to carefully read things every time I visit the page. Trying to make users who think they already know what they're doing read the text isn't really the problem. You'll get much better results by giving more points of obvious differentiation — i.e. visually distinguishing by something other than text.

@tmtmtmtm thanks for this is really great feedback. I agree that while the changes in this PR make the controls more understandable for new users, they don't do much people who've used them before. It's easy to make a slip and click the other button that looks so similar.

To help prevent these errors, I'll try visually distinguishing the buttons with colour ( as you suggested in #886 ).

equivalentideas:
I've used colour here to distinguish the buttons. I think this is working better now. @tmtmtmtm Do you think now that the buttons are different colours, and shapes, you'd avoid the error you were having?

screen shot 2016-03-29 at 10 55 03 am

jpmckinney:
What about putting a ☠ in the "Delete your scraper" button text?

equivalentideas:
Here's a quick mock of some icons, just using the inbuilt bootstrap stuff. That skull emoji is more fun :confetti_ball: but do the illustrations help?

screen shot 2016-04-05 at 10 28 26 am

tmtmtmtm:
These are definitely big improvements on what's live at the minute. 

There may be further gains to be made later, but with the icons we now have two points of differentiation, which I'd say is enough of a win to go with this and see if it helps. :shipit:

equivalentideas:
> There may be further gains to be made later, but with the icons we now have two points of differentiation, which I'd say is enough of a win to go with this and see if it helps. :shipit:

Awesome, thanks for that @tmtmtmtm ( and also @jpmckinney ).

I'll implement the icons and review the code and :shipit: if all good.

CloCkWeRX:
One idle though: Github's delete repository behaviour might be a good example to look at: "Type the name of the scraper to confirm you understand it will be deleted." if the UI changes aren't quite enough by themselves

equivalentideas:
> One idle though: Github's delete repository behaviour might be a good example to look at: "Type the name of the scraper to confirm you understand it will be deleted." if the UI changes aren't quite enough by themselves

Totally @CloCkWeRX :+1: We'll see how this goes and can try something else/more like that if the unintentional deletes continue to happen.

equivalentideas:
Deployed :ship: Thanks for your help @tmtmtmtm @jpmckinney @CloCkWeRX @henare

Issue #421, opened 2014-07-03
create tutorial


kat:
create a step by step tutorial to make it easier to get started

equivalentideas:
Seems like @emikulic has kicked this off with #796 and #798

henare:
This kind of relates to #846 too.

equivalentideas:
After our [scraping workshop last week](https://www.openaustraliafoundation.org.au/2015/08/07/introduction-to-web-scraping-workshop-in-sydney/), we got consistent feedback that it would be really helpful to have a step by step guide on how to write a scraper—exactly what this issue is calling for.

This will be a really helpful resource to workshop participants as well as people using morph generally, so I'm gonna whip something up.

equivalentideas:
Here are steps to writing a scraper, based on my own methods and what @henare demonstrated in the workshop:

1. Find the data you're looking for and work out if it can be scraped.
2. If it can be scraped, create your new scraper on morph.io. Pick the language you want to write your scraper in and add a nice description and name for the scraper so people can find it through search.
3. clone your scrape to your local machine using 'git clone scraper url'
4. make sure you have all the dependencies installed. If you're writing your scraper in Ruby, do you have Ruby installed? Do you have Bundler installed to manage all the libraries your scraper will need?
5. Now it's time to start writing your scraper.
6. open your code editor and look at the example code
7. define your object with the data you want to collect.
8. using IRB, get each piece of data for a single record (start small with just one record).
9. Once you've got each piece of data you need, consider adding a date scraped, so you can verify your data later.
10. Fill out the record you've defined in your scraper.rb file and use the 'p' (print) method to output the record you've collected when you run the scraper on the command line.
11. now add a loop to your scraper to get every record you need to on the page
12. if the records you need cover several pages, you'll need to loop through all the pages
13. save your data using the scraperwiki library
14. push your scraper to morph.io
15. run the scraper and check for errors.
16. review that data you've collected by looking at the api and downloading the csv
17. if your scraper needs to run each day, set that on morph. 
18. Celebrate!

For the tutorial, I think it would be helpful to take readers through these steps writing an example scraper. The bills in NSW parliament is a nice simple example that includes pagination, so I think I'll go with that unless we come up with something more exciting.

benrfairless:
Thanks for raising this, and sorry it has sat without a response for so long.

[...]

PR #899, opened 2015-09-08
[WIP] Add a tutorial on writing a simple scraper

equivalentideas:
Issue #421 

I've just plonked this in a markdown file while I'm writing it.

The dumb place for this to live is as another page in the documentation called “tutorial”. That's where I'm planning to park this for now. I'll have more of a think and ask for smarter ideas once there's a full draft.

### DIFF
--- a/app/controllers/examples_controller.rb
+++ b/app/controllers/examples_controller.rb
@@ -24,8 +24,10 @@ class ExamplesController < ApplicationController
       @members = @members[0..49]
     elsif @page == 2
       @members = @members[50..99]
-    else
+    elsif @page == 3
       @members = @members[100..149]
+    else
+      raise ActiveRecord::RecordNotFound
     end
     render layout: nil
   end
--- a/app/views/documentation/tutorial.md
+++ b/app/views/documentation/tutorial.md
@@ -36,9 +36,10 @@ for practice scraping.
 You will target this page to get the member information with your scraper.
 
 Some webpages are much harder to scrape than others.
-This member information is published in a simple html list,
-rather than embedded in an image.
-This means you should be able to write a scraper to collect it.
+The member information you’re trying to collect
+is published in a simple html list,
+which means you should be able to target and collect
+the information you want quite easily.
 If the information was in an image or PDF
 then it would be much harder to access programmatically and
 therefor much harder to write a scraper for.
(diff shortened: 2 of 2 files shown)


equivalentideas:
@henare could I get your feedback on the text of the  tutorial https://github.com/openaustralia/morph/blob/tutorial/app/views/documentation/tutorial.md

Anything that makes it clearer and easier to follow would be great. Flag bits that you think the reader needs a demonstration via gif or image.

There are places where I suspect it could be simplified. I want to get your thoughts on what can be cut before chopping.

Once we’re happy with the text I'll add it to the documentation section.

henare:
@equivalentideas wow, this is huge!

I'm not in the right frame of mind to look at this right now so it'll take me a few days to get back to you with feedback. In the mean time it would be good to think about where we're going to put this.

equivalentideas:
> I'm not in the right frame of mind to look at this right now so it'll take me a few days to get back to you with feedback.

No rush :+1:

> In the mean time it would be good to think about where we're going to put this.

I think as a section in the documentation could be a good spot. That way people looking for help/advice might see it.

dumping some thoughts about how we could link to this in the app :arrow_down: 

We could also include links to it in places where someone might need this kind of thing, such as when they first sign up or create a there first scraper. Lots of people sign up, create a scraper and then never edit it. This could be because they're interested, but don't know where to start. The tutorial could be one way to help them.

henare:
Any more thoughts on where to put this @equivalentideas?

We've also discussed the following issues:

* It's Ruby-only and we don't want to give the impression morph.io is Ruby-only
* It's really big and so could be overwhelming. It would be good if we could break it up

However I really like the idea of linking off to it in the application (but being Ruby-only would be an issue).

We also discussed possibly posting this as a series of blog posts.

henare:
As discussed with @equivalentideas we've decided to publish this as a blog post series. I've assigned it to me to extract posts and start posting them before we have our next scraping workshop.

henare:
https://www.openaustraliafoundation.org.au/2015/10/13/ruby-web-scraping-tutorial-on-morph-io-part-1/

henare:
https://www.openaustraliafoundation.org.au/2015/10/15/ruby-web-scraping-tutorial-on-morph-io-part-2-start-writing-your-scraper/

henare:
I've scheduled the remaining posts to go out before the end of Thursday because @equivalentideas wanted to include them in the newsletter.

Issue #905, opened 2015-09-15
Server disk full


henare:
The morph.io server has used over 180GB so the disk is 97% full. New Relic told me we had a month to deal with it but overnight the server exhausted its disk space completely.

* [x] Work out a time for the outage
* [x] Write an announcement
* [x] Send the announcement
* [x] Resize the server on Linode
* [ ] Increase the memory limit (#898)

## Work out a time for the outage

Linode estimates 2.5 hours to do the migration.

Google Analytics says in the last month 20% of our traffic has come from Australia, 16% from the UK, and 14% from the US.

## Write an announcement

We want to email everyone to tell them of the outage, thank our supporters, ask for more supporters as we'll no longer be covering our server bill, and tell them of the excellent new 5x increased memory limits (#898).

## Resize the server on Linode

Just log into the console.

## Increase the memory limit (#898)

https://github.com/openaustralia/morph/blob/master/lib/morph/docker_runner.rb#L22-L27

equivalentideas:
**9-10 am Friday morning** seems like the best times during the week based on our visitor stats in Google Analytics.

As @henare notes, this doesn't include API usage. This is something we should look into monitoring so we can better schedule these things in the future. Maybe we can see that somehow now, but I've got no idea how.

equivalentideas:
Here's a draft of an email for review @henare :

https://docs.google.com/document/d/13YELJQmtZeTn1NqlJA4qbUxzib6NTApdxOJtaFhD3xg/edit

equivalentideas:
@henare I can do the export of all users and import that into civic etc.

henare:
>9-10 am Friday morning seems like the best times during the week based on our visitor stats in Google Analytics.

:+1: great, thanks for doing this.

>@henare I can do the export of all users and import that into civic etc.

Yes, please! I'm looking at the email now.

henare:
Email is looking great, good stuff @equivalentideas!

Are you going to send that out?

equivalentideas:
> Are you going to send that out?

Yep, I'll load it and send a test and then send it out if there's no bugs.

equivalentideas:
Email sent.

henare:
```
root@li421-88:~# df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/xvda       379G  185G  191G  50% /
udev            7.9G  4.0K  7.9G   1% /dev
tmpfs           1.6G  276K  1.6G   1% /run
[...]

PR #911, opened 2015-09-17
Increase container memory limit to 512MB

### LINKED ISSUE #898: Re-consider runtime resource limits
Following on from [this discussion on the help forum](https://help.morph.io/t/why-is-my-scraper-being-killed/76/10) I would like to suggest raising the memory limit for scrapers to 512MB.

To give one example, the [FTS scraper](https://github.com/civicdataeu/fts-scraper) (EU expenditure) needs to download a zipped, 14 MB XML file. There may be constellation of stars, programming languages and streaming interfaces which makes it possible to do a streamed download of such a file, while dynamically unpacking it and then parsing the elements in the file using SAX (the XML is also broken, so we need to run a string regex first, also streaming that somehow while feeding it into the XML parser). This is just insane, makes the code brittle, and costs exponentially more time to implement than the naive solution. Still, I'd argue that processing an XML file should be within the scope of morph.io.

henare:
Fixes #898.

Can be merged and deployed after #905 is fixed.

Bleh, I accidentally pushed to the wrong branch. Hence all the stupid reverts.

### DIFF
--- a/app/views/documentation/_limits.md
+++ b/app/views/documentation/_limits.md
@@ -4,7 +4,7 @@ this.
 However, we do impose a couple of hard limits on running scrapers so they don't
 take up too many resources
 
-* max 100 MB memory
+* max 512 MB memory
 * max 24 hours run time for a single run
 
 If a scraper runs out of memory or runs too long it will get killed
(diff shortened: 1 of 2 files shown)


mlandauer:
Isn't there a reference to the size of the memory limit in the docs as well?

henare:
:+1: good thinking.

PR #937, opened 2015-10-30
Add some Perl documentation

katkad:
Please check if renders correctly. I hadn't run morph to test.


henare:
So awesome :ok_hand: Thanks so much for this great contribution! :tada:

PR #950, opened 2016-01-26
Webhooks

### LINKED ISSUE #915: Call webhook when a scraper finishes running
It would be super-useful if I could register a URL that gets automatically called when a scraper finishes running, so that I could automatically fetch the new data over the API.

Ideally this would send as a payload the information that gets displayed in the History (result, records added/removed etc), but even a simple result code would be a good first step.

chrismytton:
I've made a start on adding webhooks to scrapers. This is a work in progress but I wanted to get some feedback on the implementation before I get into refactoring and writing tests/docs.

screen shot 2016-01-26 at 10 42 27

Some thoughts:

What do we want to do about webhook failures? I know sidekiq will retry with backoff but perhaps don't want to retry at all, or only retry a couple of times or…?

When a webhook has failed several times it might be preferable to disable and/or remove the webhook to prevent it consuming resources.

I haven't added any metadata to the webhook at the moment. It just does an empty POST request to the url(s) provided. This is enough for the EveryPolitician use-case so I thought it was best to offer nothing until we had more information on what people want from webhooks. ([Requestbin example of what's delivered currently](http://requestb.in/vgo2f6vg?inspect#w5dlxs))

The interface for adding webhooks is very minimal, it would be good to do something a bit fancier in the future, at least allow people to test webhooks and see what the last response status code was for example. Perhaps this would be better on its own page rather than cluttering up the settings page further? I quite like the way https://libraries.io does its webhook settings:

screen shot 2016-01-26 at 10 47 03

I obviously need to write some tests/docs for this if it seems like a reasonable approach :grimacing: 

Fixes #915

### DIFF
--- a/app/models/run.rb
+++ b/app/models/run.rb
@@ -103,4 +103,12 @@ class Run < ActiveRecord::Base
   def env_variables
     Variable.to_hash(variables)
   end
+
+  # Called when a run has finished. Perform any post-run work here.
+  def finished!
+    return unless scraper.present?
+    scraper.update_sqlite_db_size
+    scraper.reindex
+    scraper.reload
+  end
 end
--- a/app/models/run.rb
+++ b/app/models/run.rb
@@ -106,7 +106,6 @@ class Run < ActiveRecord::Base
 
   # Called when a run has finished. Perform any post-run work here.
   def finished!
-    return unless scraper.present?
     scraper.update_sqlite_db_size
     scraper.reindex
     scraper.reload
--- a/app/models/run.rb
+++ b/app/models/run.rb
@@ -110,5 +110,6 @@ class Run < ActiveRecord::Base
     scraper.update_sqlite_db_size
     scraper.reindex
     scraper.reload
+    scraper.deliver_webhooks(self)
   end
 end
--- a/app/views/documentation/webhooks.html.haml
+++ b/app/views/documentation/webhooks.html.haml
@@ -1,0 +1,21 @@
+:markdown
+  # Webhooks
+
+  Webhooks allow you to run custom code when one of your scrapers has finished
+  running on morph.io. When a webhook is triggered, we'll send a HTTP POST
+  request to the webhook's configured URL. Webhooks can be used to transform
+  data into different formats, send out notifications, update an archive, or
+  even run some realtime analysis.
+
+  Webhooks can be installed on any of your scrapers. Once installed, they will
+  be triggered each time that scraper finishes running.
+
+  ## Payloads
+
+  Webhooks currently just make an empty HTTP POST request to the specified url.
+  If you'd like the webhooks to have some state you'll need to put that state in
+  the webhook url, either as the path or as GET parameters. If there's something
+  specific you'd like included in the webhook payload then please
+  [ask a question on our help forum](https://help.morph.io/).
+
+
--- a/app/views/scrapers/_settings.html.haml
+++ b/app/views/scrapers/_settings.html.haml
@@ -22,6 +22,9 @@
     = f.button :submit, "Update"
 
   %h3 Webhooks
+  %p
+    We'll make an HTTP POST request to each webhook url whenever this scraper finishes running.
+    = link_to 'Read documentation', webhooks_documentation_index_path
   .row
     .col-sm-3
       %label URL
--- a/app/views/scrapers/_settings.html.haml
+++ b/app/views/scrapers/_settings.html.haml
@@ -20,3 +20,24 @@
           Add variable
   - if can? :update, @scraper
     = f.button :submit, "Update"
+
+  %h3 Webhooks
+  .row
+    .col-sm-3
+      %label URL
+  #webhooks
+    = f.fields_for :webhooks do |webhook_form|
+      .nested-fields
+        .row
+          .col-sm-3
+            = webhook_form.input :url, input_html: {class: "url"}
+          .col-sm-1
+            = link_to_remove_association webhook_form do
+              %i.fa.fa-times-circle.fa-2x
+    .links
+      - if can? :update, @scraper
+        = link_to_add_association f, :webhooks do
+          %i.fa.fa-plus-circle.fa-2x
+          Add webhook
+  - if can? :update, @scraper
+    = f.button :submit, "Update"
--- a/app/views/scrapers/_settings.html.haml
+++ b/app/views/scrapers/_settings.html.haml
@@ -33,7 +33,16 @@
       .nested-fields
         .row
           .col-sm-3
-            = webhook_form.input :url, input_html: {class: "url"}
+            = webhook_form.input :url, wrapper: :vertical_input_group do
+              %span.webhook-status.input-group-addon{class: webhook_last_delivery_status(webhook_form.object)}
+                - if webhook_form.object.deliveries.any?
+                  - if webhook_form.object.last_delivery.success?
+                    %span.fa.fa-check{data: {toggle: 'tooltip'}, title: 'Last delivery was successful.'}
+                  - else
+                    %span.fa.fa-warning{data: {toggle: 'tooltip'}, title: "Last delivery was not successful. Invalid HTTP Response: #{webhook_form.object.last_delivery.response_code}."}
+                - else
+                  %span.fa.fa-plug{data: {toggle: 'tooltip'}, title: 'No deliveries yet'}
+              = webhook_form.input_field :url, class: 'form-control'
           .col-sm-1
             = link_to_remove_association webhook_form do
               %i.fa.fa-times-circle.fa-2x
--- a/app/workers/deliver_webhook_worker.rb
+++ b/app/workers/deliver_webhook_worker.rb
@@ -1,0 +1,13 @@
+class DeliverWebhookWorker
+  include Sidekiq::Worker
+  sidekiq_options backtrace: true
+
+  def perform(webhook_delivery_id)
+    webhook_delivery = WebhookDelivery.find(webhook_delivery_id)
+    res = Faraday.post(webhook_delivery.webhook.url)
+    webhook_delivery.update_attributes(
+      response_code: res.status,
+      sent_at: DateTime.now
+    )
+  end
+end
--- a/app/workers/deliver_webhook_worker.rb
+++ b/app/workers/deliver_webhook_worker.rb
@@ -4,9 +4,9 @@ class DeliverWebhookWorker
 
   def perform(webhook_delivery_id)
     webhook_delivery = WebhookDelivery.find(webhook_delivery_id)
-    res = Faraday.post(webhook_delivery.webhook.url)
+    response = Faraday.post(webhook_delivery.webhook.url)
     webhook_delivery.update_attributes(
-      response_code: res.status,
+      response_code: response.status,
       sent_at: DateTime.now
     )
   end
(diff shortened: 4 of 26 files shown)


app/models/run.rb
```diff
@@ -103,4 +103,13 @@ def variables
   def env_variables
     Variable.to_hash(variables)
   end
+
+  # Called when a run has finished. Perform any post-run work here.
+  def finished!
+    return unless scraper.present?
```
henare: Maybe move this back down to `Morph::Runner`? It made me scratch my head when I first saw it. We're also doing another check for it down on the Runner anyway.

app/workers/deliver_webhook_worker.rb
```diff
@@ -0,0 +1,13 @@
+class DeliverWebhookWorker
+  include Sidekiq::Worker
+  sidekiq_options backtrace: true
+
+  def perform(webhook_delivery_id)
+    webhook_delivery = WebhookDelivery.find(webhook_delivery_id)
+    res = Faraday.post(webhook_delivery.webhook.url)
```
henare: Resource? Result? I think a non-abbreviated variable name here would be better.

henare:
>I've made a start on adding webhooks to scrapers.

This is looking really good! I've made a couple of tiny suggestions.

>What do we want to do about webhook failures? I know sidekiq will retry with backoff but perhaps don't want to retry at all, or only retry a couple of times or…?

Just letting Sidekiq retry seems like the easy thing. Will it ever be marked as a failure though?

>When a webhook has failed several times it might be preferable to disable and/or remove the webhook to prevent it consuming resources.

I'm not too worried about the resources it consumes but we definitely need to let the user know somehow.

>I haven't added any metadata to the webhook at the moment. It just does an empty POST request to the url(s) provided. This is enough for the EveryPolitician use-case so I thought it was best to offer nothing until we had more information on what people want from webhooks.

In #915 @tmtmtmtm says...

>Ideally this would send as a payload the information that gets displayed in the History (result, records added/removed etc), but even a simple result code would be a good first step.

...which sounds like a good idea. On the other hand is this designing the API before we have a use for it?

If Sidekiq automatically retries could this be a problem if there wasn't any payload? i.e. if you include the run ID then the consuming system could detect if it was a duplicate notification.

>The interface for adding webhooks is very minimal, it would be good to do something a bit fancier in the future, at least allow people to test webhooks and see what the last response status code was for example. Perhaps this would be better on its own page rather than cluttering up the settings page further? I quite like the way https://libraries.io does its webhook settings:

What are your thoughts @equivalentideas?

>I obviously need to write some tests/docs for this if it seems like a reasonable approach :grimacing: 

:+1: 

---

I just noticed this other comment from @tmtmtmtm on https://github.com/openaustralia/morph/issues/915#issuecomment-141748922

>I'd also like to be able to do this for other people's scrapers too.

Any thoughts on that? It seems like it would be quite difficult to design an simple, understandable interface for that.

chrismytton:
> This is looking really good! I've made a couple of tiny suggestions.

@henare Thanks! I've addressed your comments and pushed some tests and documentation. Let me know of there's anything you like tested more/less/differently.


> Just letting Sidekiq retry seems like the easy thing. Will it ever be marked as a failure though?

Yes, Sidekiq's [automatic job retry](https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry) will "perform 25 retries over approximately 21 days", after which jobs go to the dead job queue. I wonder if 25 retries over 21 days is too much though?



> I'm not too worried about the resources it consumes but we definitely need to let the user know somehow.

Perhaps we could add a "latest response status code" indicator next to each webhook on the setting page for now? Then that could be expanded into something a bit more elaborate in the future if necessary.


> In #915 @tmtmtmtm says...
> 
>> Ideally this would send as a payload the information that gets displayed in the History (result, records added/removed etc), but even a simple result code would be a good first step.

> ...which sounds like a good idea. On the other hand is this designing the API before we have a use for it?

I certainly think the payloads could/should include more data, but yea I'm wary of trying to design an API without real-world use-cases. I'd prefer to get this basic implementation merged first then ticket any enhancements separately.



> If Sidekiq automatically retries could this be a problem if there wasn't any payload? i.e. if you include the run ID then the consuming system could detect if it was a duplicate notification.

Yea, including the run id in the payload might be sensible, but for the EveryPolitician use case duplicate webhooks won't have any adverse consequences. As above, I'd rather get this simple version polished and running in production and then add other things as separate enhancements :)

> I just noticed this other comment from @tmtmtmtm on #915 (comment)
> 
>> I'd also like to be able to do this for other people's scrapers too.

> Any thoughts on that? It seems like it would be quite difficult to design an simple, understandable interface for that.

Yea this would be nice but as you say, the implementation could be tricky. Once again I think this is probably best as an additional enhancement at a later stage.

---

I think there's lots of additional niceness that can be added to the webhooks, but my preference would be to ticket them and add them as future enhancements (which I'd be more than happy to do in future pull requests!) and get this running in production to gather some real world data :balloon:

henare:
Wonderful - I agree with all your points @chrismytton :ok_hand: 

>Perhaps we could add a "latest response status code" indicator next to each webhook on the setting page for now? Then that could be expanded into something a bit more elaborate in the future if necessary.

Yeah that also sounds good to me :+1: I'm assigning to @equivalentideas to see if he has any comments on the UI stuff, including this, before we merge and deploy :rocket:

app/views/documentation/webhooks.html.haml
```diff
@@ -0,0 +1,21 @@
+:markdown
+  # Webhooks
+
+  Webhooks allow you to run custom code when one of your scrapers has finished
```
equivalentideas, line 4: As someone who isn't very familiar with webhooks I find this explanation a bit confusing.

> Webhooks allow you to run custom code

When I read this I imagine that I can add some custom ruby code somewhere in morph that will run when my scraper finishes.

I think this would be clearer if it started with a basic explanation of why you might want a webhook and what it is. Something like (but simpler/shorter/clearer if possible :wink:):

> Ever wished you could be alerted whenever your morph.io scrapers finished running? Then you could send out notifications, update an archive, or run analysis as soon as new data is collected. This is what webhooks are for. You can add webhook urls to your scrapers and we'll make a HTTP POST request to it when your scraper finishes each run. You can then use this to trigger any custom code you like in an external application. Add webhook urls in your scrapers settings page.

app/views/scrapers/_settings.html.haml
```diff
@@ -20,3 +20,27 @@
           Add variable
   - if can? :update, @scraper
     = f.button :submit, "Update"
+
+  %h3 Webhooks
+  %p
+    We'll make an HTTP POST request to each webhook url whenever this scraper finishes running.
```
equivalentideas, line 26: I think this would be clearer as:

> We'll make an HTTP POST request to each webhook url every time this scraper finished a run.

equivalentideas:
> Yeah that also sounds good to me :+1: I'm assigning to @equivalentideas to see if he has any comments on the UI stuff, including this, before we merge and deploy :rocket:

Sorry for the delay all, I've just had a really full week here. This seems fantastic to me. I've made a few comments on the wording.

> Perhaps we could add a "latest response status code" indicator next to each webhook on the setting page for now? Then that could be expanded into something a bit more elaborate in the future if necessary.

I think some feedback is really essential here to help people get this setup. The status code is a good start, but it seems to me that it's not very useful if you don't recognise it as a status code or don't know which status codes mean what. A lot of developers using morph.io will know this, but there are people who know. Do you think it would be helpful to complement the status code with a colour to say whether it was successful or not, and even an icon? Here’s Github’s solution to this:

screen shot 2016-02-05 at 5 01 28 pm

Also some validation on the url format will save users the frustration of unnecessary debugging runs and errors in submissions if they have typos or assume the wrong url format.

Aside from that this all looks simple and awesome. Can't wait to see it shipped!

henare:
@equivalentideas I'm not clear on what you think must be done before we merge this. Can you make a list so we can get through the items and get this deployed?

As I said earlier, I agree with @chrismytton:

>I think there's lots of additional niceness that can be added to the webhooks, but my preference would be to ticket them and add them as future enhancements (which I'd be more than happy to do in future pull requests!) and get this running in production to gather some real world data :balloon:

equivalentideas:
>  Can you make a list so we can get through the items and get this deployed?

* [ ] [make the documentation wording clearer](https://github.com/openaustralia/morph/pull/950#discussion_r51980069)
* [ ] [give users a feedback on whether their webhooks are working](https://github.com/openaustralia/morph/pull/950#issuecomment-180216255)

I think those two bits are important for basic usability and should be completed before we merge this.

For me, the wording on the settings page is a smaller issue that we can iterate on once merged. I think the validation on the webhook url can also be done after merging.

henare:
Thanks @equivalentideas. Are you OK to make those changes @chrismytton?

chrismytton:
@henare @equivalentideas Sorry only just saw this! Yep, happy to make those changes, I'll do it first thing Tuesday UK time if that's cool? :+1:

henare:
@chrismytton excellent, no problem at all! Sorry it's taken longer than usual to get this merged and deployed, we've been in the middle of our annual planning and first round OGP planning.

chrismytton:
I've now added an icon to indicate delivery status:

### Success

screen shot 2016-02-09 at 12 04 14

### No deliveries

screen shot 2016-02-09 at 12 04 30

### Failure

screen shot 2016-02-09 at 12 04 21

Still need to update the docs, will try to get to that later today :+1:

equivalentideas:
Nice one @chrismytton that's looking really good :+1:

Could you explain the thinking behind the 'no deliveries' icon? We could alternatively have no icon for that state, though I do think the text “no deliveries yet” is helpful.

chrismytton:
@equivalentideas My thinking was that webhooks are like plugging web apps into each other, and if there have been no deliveries then the plug is unplugged. A bit obscure?! No icon might be better. The other option would be to send an initial "ping" webhook request so that we can show some kind of item almost immediately and show the user if their webhook is working or not?

equivalentideas:
> The other option would be to send an initial "ping" webhook request so that we can show some kind of item almost immediately and show the user if their webhook is working or not?

I like the idea to provide feedback on the input immediately so people can correct any errors—but that sounds like something to pick up in a ticket. I think for this first pass it’ll be clearer to have no icon for the “no deliveries” state.

henare:
Oh no, it looks like this pull request got stuck! :frog: 

I've once again reviewed where it's at and I'm comfortable merging :shipit: 

I think the only bit of outstanding feedback was:

>I think for this first pass it’ll be clearer to have no icon for the “no deliveries” state.

I suspect that feedback might not have been based on testing this in place. I say this because while I agree with the feedback in principle, actually using this it seems fine as is.

henare:
Deployed :rocket: Great stuff @chrismytton, thank you! :tada: 

Now, who wants to write a blog post about this fancy new feature? :wink:

chrismytton:
:sparkles: Awesome! :sparkles: Thanks @henare! I'll open issues for the other things we discussed in this thread when I get a chance later.

henare:
Sounds great @chrismytton.

Actually my cheeky blog post idea above might not have been that cheeky. I think it would be really interesting to hear how @chrismytton and @tmtmtmtm use this in the real world with EveryPolitician.

That kind of practical example is just the thing to inspire other people to come up with new and interesting ways to use a feature like this. Once you start using it if you do feel like writing that up, we'd definitely be keen to cross post it on the OAF blog.

PR #958, opened 2016-02-23
Slightly improve webhooks settings helper text

equivalentideas:
This is a small improvement to make the webhooks settings helper text a little clearer.

It's currently:

> We'll make an HTTP POST request to each webhook url whenever this scraper finishes running.

The wording “whenever this scraper finishes running”, uses longer, ambiguous words and could mean “when this scraper finally finishes running”. If you have a prior knowledge of morph.io you can logically interpret the intended meaning, but there is no need for us to require this knowledge.

These commits change it to:

> We’ll make an HTTP POST request to each webhook url every time this scraper finishes a run.

This is more specific and uses simpler, unambiguous words, so people can get all the knowledge they need directly from the text.

screen shot 2016-02-23 at 12 00 19 pm

### DIFF
--- a/app/views/scrapers/_settings.html.haml
+++ b/app/views/scrapers/_settings.html.haml
@@ -23,7 +23,7 @@
 
   %h3 Webhooks
   %p
-    We’ll make an HTTP POST request to each webhook URL every time this scraper finishes a run.
+    We’ll make an HTTP POST request to each webhook URL every time this scraper finishes running.
     = link_to 'Read documentation', webhooks_documentation_index_path
   .row
     .col-sm-3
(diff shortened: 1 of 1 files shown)


henare:
Slight changes indeed! :smiley: I'll comment on them anyway and merge after a similarly slight tweak.

>The wording “whenever this scraper finishes running”, uses longer, ambiguous words and could mean “when this scraper finally finishes running”.

I like that change and think you're right that it's slightly ambiguous and could mean what you say above.

>If you have a prior knowledge of morph.io you can logically interpret the intended meaning, but there is no need for us to require this knowledge.

I like the sentiment here and that's why I'm perplexed about the other change you've made but not mentioned. A program "finishes running" is commonly used but "finishes a run" is using morph.io's language.

PR #969, opened 2016-03-16
Randomise list of heroes; closes #893

### LINKED ISSUE #893: Randomise sort of heroes on front page
Right now it looks to be sorted by sign up date.

snoblenet:
This pull request randomises the Owner.supports(plan) scope so heroes are listed in random order. Note, this change uses an MySQL only option, but as Morph.io seems pretty closely bound with MySQL, that seemed fine.

Note: also loops in a minor change I made to the wording of the .gitignore file before I made a branch for this feature.

closes #893

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@
 /tmp/sessions
 /tmp/sockets
 
-# Ignore the PID tempfiles but not the folder that holds them.
+# Ignore the PID tempfiles but not PID folder that holds them.
 /tmp/pids/*.pid
 
 /.env
--- a/app/models/owner.rb
+++ b/app/models/owner.rb
@@ -18,7 +18,7 @@ class Owner < ActiveRecord::Base
   serialize :feature_switches
 
   # Supporters that are on a particular plan
-  scope :supporters, ->(plan) { where(stripe_plan_id: plan.stripe_plan_id).order(order: "RAND()") }
+  scope :supporters, ->(plan) { where(stripe_plan_id: plan.stripe_plan_id).order("RAND()") }
 
   scope :all_supporters, -> { where.not(stripe_plan_id: "") }
 
--- a/app/models/owner.rb
+++ b/app/models/owner.rb
@@ -18,7 +18,7 @@ class Owner < ActiveRecord::Base
   serialize :feature_switches
 
   # Supporters that are on a particular plan
-  scope :supporters, ->(plan) { where(stripe_plan_id: plan.stripe_plan_id) }
+  scope :supporters, ->(plan) { where(stripe_plan_id: plan.stripe_plan_id).order(order: "RAND()") }
 
   scope :all_supporters, -> { where.not(stripe_plan_id: "") }
 


snoblenet:
Don't merge yet please -- needs further work

snoblenet:
The new work is done in commit [d1de84b](https://github.com/openaustralia/morph/pull/969/commits/d1de84b55b2a0c7448ccce4a34e87b323bb228cc).

.gitignore
```diff
@@ -18,7 +18,7 @@
 /tmp/sessions
 /tmp/sockets
 
-# Ignore the PID tempfiles but not the folder that holds them.
+# Ignore the PID tempfiles but not PID folder that holds them.
```
henare, line 21: Is this supposed to be part of this pull request?

app/models/owner.rb
```diff
@@ -18,7 +18,7 @@ class Owner < ActiveRecord::Base
   serialize :feature_switches
 
   # Supporters that are on a particular plan
-  scope :supporters, ->(plan) { where(stripe_plan_id: plan.stripe_plan_id) }
+  scope :supporters, ->(plan) { where(stripe_plan_id: plan.stripe_plan_id).order("RAND()") }
```
henare, line 21: This is effectively setting a default order of random on this scope. This feels like it could become very confusing because it's easy to miss - you'd be wondering why you were getting different results each time you call this. This is the same reason [default scopes are to be avoided](http://www.databasically.com/2011/02/07/default_scope-use-cases-caveats-and-work-arounds/).

For this reason I'd suggest doing this at the controller level. What do you think?

henare:
Woo :tada: great to have you working on this @snoblenet. I've made a couple of comments, over to you.

app/models/owner.rb
henare, line 21: > Note, this change uses an MySQL only option, but as Morph.io seems pretty closely bound with MySQL, that seemed fine.

If you do decide to make the change I've suggested above, I'd also recommend doing it in Ruby since it will be easier to read. Performance won't be an issue so clarity should trump that in this case :sparkles:

snoblenet:
At present, the logic is in the view in a form that cannot easily be moved to the controller.

In `/app/views/static/_featured_supporters.html.haml`:

```
- Plan.featured_plans.each do |plan|
  - Owner.supporters(plan).each do |owner|
     %li.owner-item= link_to owner_image(owner, 60), owner

```
To address the current issue, I'd recommend updating the view to read:

````
- Plan.featured_plans.each do |plan|
  - Owner.supporters(plan).shuffle.each do |owner|
     %li.owner-item= link_to owner_image(owner, 60), owner
````

After that, I think it's worth opening a new branch to remove logic from the view. I'd be inclined to move this sort of logic to a view helper rather than the controller. I'd be happy to take that on. Just let me know.

henare:
>After that, I think it's worth opening a new branch to remove logic from the view. I'd be inclined to move this sort of logic to a view helper rather than the controller. I'd be happy to take that on. Just let me know.

:+1: to both those suggestions.

PR #979, opened 2016-03-22
Show how long scraper has been erroring

### LINKED ISSUE #320: Say how long a scraper has been erroring on list pages
e.g. on this page you can't tell how long scrapers have been broken for: https://morph.io/planningalerts-scrapers

snoblenet:
Closes #320
### Before

selection_002
### After

selection_003

### DIFF
--- a/app/controllers/owners_controller.rb
+++ b/app/controllers/owners_controller.rb
@@ -10,19 +10,8 @@ class OwnersController < ApplicationController
     # Only do this once
     session[:new_supporter] = false if @new_supporter
 
-    @scrapers = @owner.scrapers.includes(:last_run)
+    @scrapers = @owner.scrapers.order(:updated_at).includes(:last_run)
 
-    # Split out scrapers into different groups
-    @running_scrapers, @erroring_scrapers, @other_scrapers = [], [], []
-    @scrapers.each do |scraper|
-      if scraper.running?
-        @running_scrapers << scraper
-      elsif scraper.requires_attention?
-        @erroring_scrapers << scraper
-      else
-        @other_scrapers << scraper
-      end
-    end
   end
 
   def settings_redirect
--- a/spec/features/scrapers_index_spec.rb
+++ b/spec/features/scrapers_index_spec.rb
@@ -11,24 +11,51 @@ describe "scraper exploration" do
       @unique_on_github_1 = create :scraper,
         :owner => @organization,
         :name => "unique_on_github_1",
-        :description => "Unique on GitHub 1"
+        :description => "Unique on GitHub 1",
+        :auto_run => true
       @unique_on_github_2 = create :scraper,
         :owner => @organization,
         :name => "unique_on_github_2",
-        :description => "Unique on GitHub 2"
+        :description => "Unique on GitHub 2",
+        :auto_run => true
+      @unique_on_github_3 = create :scraper,
+        :owner => @organization,
+        :name => "unique_on_github_3",
+        :description => "Unique on GitHub 3",
+        :auto_run => true
     end
 
-    @last_run = create :run,
+    @first_run_for_scraper_1 = create :run,
       :scraper => @unique_on_github_1,
-      :finished_at => Time.now - (3 * 24 * 60 * 60),
+      :finished_at => Time.now - 230000,      
+      :queued_at => Time.now - 210000,
       :status_code => 0
               
+    @last_run_for_scraper_1 = create :run,
+      :scraper => @unique_on_github_1,
+      :finished_at => Time.now - 3000,
+      :queued_at => Time.now - 1000,
+      :status_code => 1
+              
+    @last_run_for_scraper_3 = create :run,
+      :scraper => @unique_on_github_3,
+      :finished_at => Time.now - 3000,
+      :queued_at => Time.now - 1000,
+      :status_code => 0
+
     visit "/planningalerts-scrapers"
-    within("div#unique_on_github_1") do
-      expect(page).to have_content "Errored for 3 days"
-    end
-    within("div#unique_on_github_2") do
-      expect(page).to_not have_content "Errored"
+    save_and_open_page
+    within("div.scraper-alerts-list") do
+      within("div#unique_on_github_1") do
+        expect(page).to have_content "Errored for 3 days"
+      end
+      within("div#unique_on_github_2") do
+        expect(page).to have_content "Never run successfully"
+      end
+      within("div#unique_on_github_3") do
+        expect(page).to_not have_content "Errored for"
+        expect(page).to_not have_content "Never run successfully"  
+      end
     end
   end
 
--- a/spec/features/scrapers_index_spec.rb
+++ b/spec/features/scrapers_index_spec.rb
@@ -44,7 +44,6 @@ describe "scraper exploration" do
       :status_code => 0
 
     visit "/planningalerts-scrapers"
-    save_and_open_page
     within("div.scraper-alerts-list") do
       within("div#unique_on_github_1") do
         expect(page).to have_content "Errored for 3 days"
--- a/spec/features/scrapers_index_spec.rb
+++ b/spec/features/scrapers_index_spec.rb
@@ -1,0 +1,35 @@
+require 'spec_helper'
+
+describe "scraper exploration" do
+
+  it "tells you how long a scraper has been erroring" do
+
+    @organization = create :organization,
+      :nickname => "planningalerts-scrapers"
+
+    VCR.use_cassette('create_valid_scraper', allow_playback_repeats: true) do
+      @unique_on_github_1 = create :scraper,
+        :owner => @organization,
+        :name => "unique_on_github_1",
+        :description => "Unique on GitHub 1"
+      @unique_on_github_2 = create :scraper,
+        :owner => @organization,
+        :name => "unique_on_github_2",
+        :description => "Unique on GitHub 2"
+    end
+
+    @last_run = create :run,
+      :scraper => @unique_on_github_1,
+      :finished_at => Time.now - (3 * 24 * 60 * 60),
+      :status_code => 0
+              
+    visit "/planningalerts-scrapers"
+    within("div#unique_on_github_1") do
+      expect(page).to have_content "Errored for 3 days"
+    end
+    within("div#unique_on_github_2") do
+      expect(page).to_not have_content "Errored"
+    end
+  end
+
+end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -9,6 +9,7 @@ ENV['RAILS_ENV'] ||= 'test'
 require File.expand_path('../../config/environment', __FILE__)
 require 'rspec/rails'
 require 'capybara/rspec'
+require 'capybara/rails'
 
 # Commented out for the benefit of zeus
 # require 'rspec/autorun'
@@ -88,7 +89,11 @@ RSpec.configure do |config|
       ex.run
     end
   end
-
+  
+  config.after do |example|
+    save_and_open_page if example.metadata[:type] == :feature and example.exception.present?
+  end
+  
   config.filter_run_excluding docker: true if ENV['DONT_RUN_DOCKER_TESTS']
   config.filter_run_excluding slow: true unless ENV['RUN_SLOW_TESTS']
 
@@ -97,3 +102,5 @@ RSpec.configure do |config|
       Sidekiq::Worker.clear_all
     end
 end
+
+Capybara.asset_host = "http://localhost:3000"
--- a/.gitignore
+++ b/.gitignore
@@ -17,10 +17,12 @@
 /tmp/miniprofiler
 /tmp/sessions
 /tmp/sockets
+/tmp/capybara
 
 # Ignore the PID tempfiles but not the folder that holds them.
 /tmp/pids/*.pid
 
+/public
 /.env
 /.env.local
 /.env.production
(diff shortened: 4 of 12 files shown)


henare:
@snoblenet nice one, this is a big thing to add :rocket: so great to have tests - that really makes me feel more confident when we come to merging.

I've added some before and after screenshots :point_up: to make this easy to review.

In testing this I've noticed these issues that need resolving:

- [ ] This isn't shown on All Scrapers page. Is there a reason we shouldn't show it?

selection_001

- [ ] In my development scrapers on an org are shown as having errored for _X_ days when they ran *successfully* that long ago:

selection_002

selection_003

Weirdly one shows on my page OK:

selection_001

snoblenet:
Ah, sorry, I created a 'before' and 'after' screenshot but I forgot to upload them. Never mind -- given I'm not working with whatever factory data I write into the tests, my screenshots are a lot less pretty than yours anyway.

I didn't look at the "All scrapers" page as it wasn't described in the original issue. However, I can include it. I'll do that now, along with addressing the other issue that you mention.

snoblenet:
@henare Regarding your "All scrapers" page question, let me answer the question with a question: Is there any situation where (1) the site is listing some scrapers, and (2) at least one of the listed scrapers is erroring, and (3) we **_do not_** want the fact that it is erroring to be revealed?

Currently, we're passing a `show_errored_detail` local to the `scrapers/scraper_base` partial. If the answer to my question is no, then we can dispense with it, and update the `scrapers/scraper_base` partial from this...

      - if show_errored_detail
        .scraper-error-info.row
          - if scraper.latest_successful_run_time
            Errored for #{time_ago_in_words(scraper.latest_successful_run_time)}
          - else
            Never run successfully

...to this...

      - if scraper.requires_attention?
        .scraper-error-info.row
          - if scraper.latest_successful_run_time
            Errored for #{time_ago_in_words(scraper.latest_successful_run_time)}
          - else
            Never run successfully

This would simplify the code, address your question about the display of errors for scrapers that are running successfully, and ensure errors are always visible. 

On the other hand, if there are situations were we want to hide the error notice, please let me know and I'll write tests for all contingencies.

henare:
@snoblenet I don't know the answers to your questions - I'd just be looking through the code and application same as you :smile: 

I think @equivalentideas may have originally wrote some of this, maybe he remembers how it works?

snoblenet:
Sure that would be great. In the mean time I will update the pull request based around the one page that was originally mentioned in the issue and the concerns that have been raised about the behaviour of`the `.scraper-error-info` div. If I get that to the point where it's ready to merge we can open a new issue for any other pages later.

henare:
>If I get that to the point where it's ready to merge we can open a new issue for any other pages later.

:+1: good call.

equivalentideas:
>  Is there any situation where (1) the site is listing some scrapers, and (2) at least one of the listed scrapers is erroring, and (3) we do not want the fact that it is erroring to be revealed?

> I think @equivalentideas may have originally wrote some of this, maybe he remembers how it works?

We decided not to show whether the scraper is errored on the scraper search results page, and I think we should stick to that for now.

> In the mean time I will update the pull request based around the one page that was originally mentioned in the issue

That sounds right to me :+1:

snoblenet:
I've simplified the controller and the view as there was no longer any need to break @scrapers into three different collections. FYI, this creates an opportunity to replace the only call to the one-line `scraper_without_owner` partial with the contents of that partial.

snoblenet:
BTW, now that we've made these changes to the markup, we could further simplify the CSS a little and create further consistency between the scraper listings if the currently working scrapers said something like "Successfully run two days" ago where the others say "Errored for two days", in the same font etc but with grey text and a subtle grey background. Thoughts?

equivalentideas:
> create further consistency between the scraper listings if the currently working scrapers said something like "Successfully run two days" ago where the others say "Errored for two days"

While consistency is helpful when people are looking to distinguish variables within scrapers, we added this panel to help them keep scrapers running by making errored scrapers more noticeable. So I don't think it would help people to add this bottom panel layout on non-errored scrapers for the moment. It would make the errored ones stand out less. Also I'm not sure the information of when a scraper passed is useful when browsing. As always, <chant>what is the user need?</chant> :guitar: 

I'll review these changes now.

spec/spec_helper.rb
```diff
@@ -88,7 +89,11 @@
       ex.run
     end
   end
-
+  
+  config.after do |example|
+    save_and_open_page if example.metadata[:type] == :feature and example.exception.present?
```
equivalentideas, line 94: I'm pretty focused on the test results, rather than wanting to see the page every time there's a failure.

Is there a way to enable this feature so that people can use it if they want? For example, if you were working on an error and couldn't understand the problem from the test output, you could add this to the test yourself?

Having it on all the time in the app seems a bit prescriptive to me—mainly because it takes my focus to a different app when I want to look at the test failure message.

Thoughts @snoblenet ?

spec/features/scrapers_index_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      :status_code => 1
+              
+    @last_run_for_scraper_3 = create :run,
+      :scraper => @unique_on_github_3,
+      :finished_at => Time.now - 3000,
+      :queued_at => Time.now - 1000,
+      :status_code => 0
+
+    visit "/planningalerts-scrapers"
+    within("div.scraper-alerts-list") do
```
equivalentideas, line 47: I think this parent 'within' call just tests that the scraper items are inside this HTML element, which doesn't seem useful? We could remove because the scraper element are targeted by `id`.

spec/features/scrapers_index_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      within("div#unique_on_github_1") do
```
equivalentideas, line 48: I don't think we care if it's a `<div>` and `<li>` that the `id` is on. @snoblenet do you think we should strip this back to `within("#unique_on_github_1")`, removing `div`?

spec/spec_helper.rb
snoblenet, line 94: FYI it only activates in feature tests, where you'd certainly want to see the page. It doesn't activate in unit tests, which represent the overwhelming majority of the app's tests. Does that change your view @equivalentideas ?

app/controllers/owners_controller.rb
```diff
@@ -10,19 +10,8 @@ def show
...
-      if scraper.running?
-        @running_scrapers << scraper
-      elsif scraper.requires_attention?
-        @erroring_scrapers << scraper
-      else
-        @other_scrapers << scraper
-      end
-    end
+    @scrapers = @owner.scrapers.order(:updated_at).includes(:last_run)
+
```
equivalentideas, line 14: Is this stray whitespace as compared to the style in used in the app @snoblenet https://github.com/openaustralia/morph/blob/master/app/models/connection_log.rb#L13-L14

I don't think we have empty lines at the end of blocks.

snoblenet, line 14: It's possible I accidentally introduced an extra line return.

spec/features/scrapers_index_spec.rb
snoblenet, line 48: @equivalentideas Sure. It's a hangover from testing the view back when some scrapers were intentionally displaced inside that div and others were intentionally displayed inside another div. It's not required now.

app/controllers/owners_controller.rb
```diff
@@ -10,19 +10,8 @@ def show
...
-    @scrapers.each do |scraper|
-      if scraper.running?
-        @running_scrapers << scraper
-      elsif scraper.requires_attention?
-        @erroring_scrapers << scraper
-      else
-        @other_scrapers << scraper
-      end
-    end
+    @scrapers = @owner.scrapers.order(:updated_at).includes(:last_run)
```
equivalentideas, line 13: These lists were divided in https://github.com/openaustralia/morph/commit/4afea5d3fe683f03ceb6969d14b164849154465a and https://github.com/openaustralia/morph/commit/4bd3911351144b65cbc8b0a7ce9cabfacae06ef0 to make the scrapers people are most likely interested in more prominent. @snoblenet could you explain how merging the lists relates to this issue and helps users?

spec/spec_helper.rb
equivalentideas, line 94: >  Does that change your view @equivalentideas ?

Not really ;-)

> where you'd certainly want to see the page.

I'm always most focused on the test error, but maybe I should look at the page more. I'm happy to try this and we can always remove it if it gets in the way :+1:

spec/features/scrapers_index_spec.rb
equivalentideas, line 48: Nice :+1: I recon it should be removed then, will only be confusing in the future.

app/controllers/owners_controller.rb
snoblenet, line 13: It's not so much that change was required by the issue. It's more that once I implemented the changes that were required by the issue, we were cycling through three object collections (@running_scrapers, @erroring_scrapers, @other_scrapers) in three identical ways in the one view. If you wanted to list the non-erroring scrapers first we could still achieve this with a sort on the one object collection (@scrapers).

spec/features/scrapers_index_spec.rb
snoblenet, line 48: Sure. But for now, I'll sit tight until we resolve our other conversation, because if we go back to having three lists of scrapers then we'll want to check the right scrapers are in each list.

app/controllers/owners_controller.rb
equivalentideas, line 14: Nice, there's a bit of stray whitespace in some of the other changes than should be stripped :mag:

app/controllers/owners_controller.rb
equivalentideas, line 13: I think you're taking two steps at once. The PR should just address the issue and then additional changes can be made in new PRs, and discussed in new issues. Focused PRs are a million times easier to view cause you've already had most of the discussion on an issue. I think you should revert it back to how it was and then open a new issue about merging the lists if you think it will be better for users.

equivalentideas:
@snoblenet I'm excited to see this happening, I think it'll be a helpful change for people and will prompt them to fix some scrapers :dancer: 

Can you address/make the changes I've suggested in the comments I've made and also make sure you've addressed this [important comment](https://github.com/openaustralia/morph/pull/979#issuecomment-199733093) by @henare above:

>  In my development scrapers on an org are shown as having errored for X days when they ran successfully that long ago:

I can see that bug on my setup too.

spec/features/scrapers_index_spec.rb
equivalentideas, line 48: :+1:

snoblenet:
Hi @equivalentideas @henare. I've just reviewed this issues page as a whole. In summary, I think there are three main themes:

1. Ensure this is the smallest possible change that can fully address the issue
2. Style issues such as line returns after blocks
3. Ensure non-erroring scrapers do not appear to be erroring

To address the first of these themes, I'll submit a new, slimmer PR as soon as I'm able to carve out sufficient time. That all makes sense. Thanks for the advice.

To address the second of these themes, should I start work on a wiki page about code style for this project?

Regarding the third theme, my feature test creates both erroring and non-erroring scrapers and checks whether each of these is displaying as it should. You can confirm this by inserting `save_and_open_page` into the test after the page is visited and then running the test. However, non-erroring scrapers in your local copy of the production database are still showing as erroring. This mean the current feature test is not sufficient. To change the feature test so that it is sufficient, we have a few options:

1. You allow me to work with a local copy of the production database, and I use this to both manually test my branch before I submit my PR and to update `@last_run_for_scraper_1`, `@unique_on_github_1`, etc to address the issue in question
2. You go into your local copy of the production data, send me a HIRB table or similar for a non-erroring scraper that is currently showing as errored, and send me a HIRB table or similar for each of the scraper's runs and other associated objects, and I use this to update update `@last_run_for_scraper_1`, `@unique_on_github_1`, etc to address the issue in question
3. You examine your local copy of the production data and then advise me how I should update `@last_run_for_scraper_1`, `@unique_on_github_1`, etc to address the issue in question

My preference would be option 1. https://github.com/sunitparekh/data-anonymization may help with this.

equivalentideas:
Thanks @snoblenet :+1: 

> To address the first of these themes, I'll submit a new, slimmer PR as soon as I'm able to carve out sufficient time. That all makes sense. Thanks for the advice.

Thanks @snoblenet I think it would be better to revert the commits with extra changes on this PR. That way we keep all this discussion.

> To address the second of these themes, should I start work on a wiki page about code style for this project?

I was thinking that something like this could be good myself. But we would prefer something that was part of the code, like a slim setup of [RuboCop](https://github.com/bbatsov/rubocop#vim) for some of our basic styles. We could use that locally and also have [Hound](https://houndci.com/) test PRs maybe. That way contributors don't have to go read some docs, they can just get comments direct on a PR or get feedback in their IDE if they have linting configured.

> Regarding the third theme

I've got to think a bit more about this one and get back to you. In the mean time I'd say try and generate a bunch of data locally by adding and running, killing scrapers locally. That's what both our data would be made up of.

mlandauer:
@equivalentideas @henare @snoblenet where is this PR at these days? Sorry to but in to the end of what's been a very long discussion and ask a dumb question.

snoblenet:
sorry @mlandauer -- i'm really happy for you to merge the PR if everyone else is happy but I can't see myself having the time to further modify the branch if it's not considered ready to merge

henare:
It sounds like @equivalentideas still has some outstanding changes requested on this and @snoblenet is unlikely to have time to work on this so I'm going to close this for now. Thanks for everyone's efforts on this, hopefully someone else will get the time to resurrect it and use the work already done here :hammer_and_wrench:

Issue #919, opened 2015-09-22
Consider limiting the number of lines of output your scraper can write


henare:
We currently have the equivalent of an unlimited scrollback buffer. As it's the cause of many problems, such as #612, I think it could be sensible to put a limit on it.

Generally if you're outputting a crazy number of lines you've stuffed something up and want to stop your scraper and fix it anyway.

henare:
One Passenger process was taking up 10GB just now. I'm pretty sure it's when it tries to respond to a request that has ridiculous amounts of log lines.

henare:
I just checked and a problematic scraper we have currently has 455,837 lines. Another scraper running right now that has lots but seems to be performing OK has 15,019 lines.

Maybe quarter of a million lines is a good compromise?

henare:
Another problematic one has 707,443 lines.

henare:
The problematic ones this morning appear to be https://morph.io/blablupcom/tyru and https://github.com/blablupcom/ntreyt - these both have 95,000 lines but the pages load OK (i.e. the log lines don't seem to be massive.

henare:
We recently had a bunch of scrapers that were infinite loops, spitting out lines during each loop. In this case the scraper can merrily run for 24 hours, outputting a several hundred thousand (so far) lines. The container only runs for 24 hours but then the sidekiq process has to sit there for a lot longer extracting all the log lines. This takes up lots of CPU and available sidekiq job spots.

mlandauer:
In #990 the approach taken was to kill a scraper that had too many lines (whatever we decided that would be). As mentioned in the same PR another approach is to limit the number of lines in the console output and not kill the scraper. There would also obviously need to be some message to the user that the console output has been limited.

The advantage of the second approach is that it is less upsetting to users. They still have the option of stopping the scrape themselves but they know at least that it's continuing to run even if they can't see the output.

mlandauer:
I'm going to have a go at implementing this by putting either a wrapper process around the scraper or by piping the output of the scraper through another little script. We'll see what works.

mlandauer:
Why does this seem to happen on a regular basis? You start working through the solution. Get stuck. Make some progress. Then, have a first working piece. Then, only then do you realise that there's another way of solving the problem. Grrr...

Another way to solve this is to put the limiting of the number of log lines in at the reading of the log lines stage. If we just don't pass any further lines up the chain then it shouldn't hog the resources. We just need to make sure that there aren't some major performance penalties to this approach.

[...]

PR #990, opened 2016-04-19
Limit scraper log lines to 50,000

henare:
This is a hacked together attempt at #919. It's not tested and doubtless
has plenty of problems.

Further caution: written while grumpy about a scraper backlog on
morph.io :rage:

This needs to be reviewed, tested, refined, and tests added.

### DIFF
--- a/lib/morph/docker_runner.rb
+++ b/lib/morph/docker_runner.rb
@@ -108,10 +108,16 @@ module Morph
           Morph::DockerUtils.container_with_interactive_connection(
             container, read_timeout: 5.minutes)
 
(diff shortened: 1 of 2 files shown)


mlandauer:
This PR kills a job if it goes over a certain number of log lines. An alternative approach would be to somehow limit the number of log lines without actually killing the job. It would obviously need to still alert the user of this

henare:
Let's close this since we're not actively working to merge this.

mlandauer:
Hmmm. I think it's worth keeping open for the time being.

mlandauer:
Given that I'm working on a different approach it does seem to make sense to me to close this now

PR #1012, opened 2016-07-02
Upgrade excon gem to fix problem when accessing docker 

mlandauer:
The problem only occurs when accessing docker through a local socket. Before on OS X I was using a network connection to docker so wasn't seeing the issue. This does the minimal upgrade to fix the problem.

if we were to upgrade to the absolute latest version of `excon` that would cause problems with the current version of `docker-api`. So, best to seperate these changes and do the smallest thing now


henare:
Travis is :green_apple: so merging.

PR #1014, opened 2016-07-04
Refactor morph database class

mlandauer:
Extract out a new class for just the database diffing part and then refactor the most ugly and complicated methods to hopefully be simpler to understand. Used codeclimate as a guide.


lib/morph/sqlite_diff.rb
```diff
@@ -6,75 +6,79 @@ def self.diffstat_safe(file1, file2)
...
+        result[:tables][:unchanged].count
+
+      result
+    end
 
-          return result
+    def self.diffstat(file1, file2)
+      SQLite3::Database.new(file1) do |db1|
+        SQLite3::Database.new(file2) do |db2|
+          return diffstat_db(db1, db2)
```
henare, line 77: Do you think it's more consistent if we implicitly return here?

lib/morph/sqlite_diff.rb
```diff
@@ -106,6 +110,12 @@ def self.diffstat_table(table, db1, db2, page = 1000)
         min = [min1, min2].min
         max = [max1, max2].max
       end
+      [min, max]
+    end
+
+    # Page is the maximum number of records that are read into memory at once
+    def self.diffstat_table(table, db1, db2, page = 1000)
```
henare, line 117: Rather than describing the parameter in a comment I was trying to think of a better name but I can't come up with one :(

henare:
Looks like lots of nice changes here :+1: I'll assign it back to you for merge and deploy @mlandauer :rocket:

mlandauer:
Scary. Doing my first deploy in a *long* time. Let's see how much I can screw this up! ;-)

PR #1018, opened 2016-07-15
Renew help.morph.io SSL certificate

henare:
I deployed this but OCSP servers don't have the certificate yet so every browser but Chrome fails. On Firefox you just get a `SEC_ERROR_OCSP_UNKNOWN_CERT` error and can't continue.

A year ago or so this started happening with StartSSL certificates so I'd just issue them then install them 24 hours later. Then a while back they seemed to have fixed the problem so I didn't worry this time but I guess it's back :disappointed: 

I think we might be able to [check it here](https://certificate.revocationcheck.com/check) before doing another provision. Currently under "OCSP response information" it says, "Status: Unknown", which sounds like a problem.


henare:
This has been provisioned.

PR #1023, opened 2016-07-18
Update gems

mlandauer:
Does a whole pile of updates to gems to get them to recent versions.

There are still a few (3) locked gems. For the time being we're sticking to Rails 4 and there are a couple of gems that would require Ruby to be updated (which we should do!) for them to be updated.


mlandauer:
Before deploying these changes it would be safest to do #1026 first

henare:
I've given this a smoke test and it looks good to me.

PR #1024, opened 2016-07-19
Update ruby to 2.3.1

mlandauer:
Update development to ruby 2.3.1. Also installs ruby 2.3.1 in production.

Note that the ansible provisioning part of this is untested as I currently don't have enough disk space and memory on my laptop to handle this. :-(

Please merge PR https://github.com/openaustralia/morph/pull/1023 before merging this one assuming you're all happy with it, of course.

Thanks!


henare:
I've tested this merged into #1023 and it looks good to me.

PR #1025, opened 2016-07-19
Update final gems

mlandauer:
Have a look at this after merging https://github.com/openaustralia/morph/pull/1024

This does another update of gems that needed later versions of ruby.

So, that means that with the exception of rails everything should be at the latest version.


henare:
I've tested this merged and it looks good to me.

PR #1032, opened 2016-07-22
Rake task to remove log lines for old runs (not the most recent)

### LINKED ISSUE #999: Purge old log lines
Our log lines table is a little large:

```
> LogLine.count
 => 99230256 
> 
```

Since only the output of the last run is available, can we just purge all log lines for runs that aren't the most recent?

mlandauer:
This fixes #999.

I've tested it locally on my development box that's seeded with old production data and it looks to work fine. I've checked a few scrapers and they all have their logs visible in the UI. I've also done a count of the log lines and they're massively reduced in number after running this.

I'd like to run this in production as soon as possible.

### DIFF
--- a/lib/tasks/app.rake
+++ b/lib/tasks/app.rake
@@ -123,6 +123,18 @@ namespace :app do
     end
   end
 
+  desc "Remove log lines for old runs (not the latest ones)"
+  task clean_up_old_log_lines: :environment do
+    Scraper.all.each do |scraper|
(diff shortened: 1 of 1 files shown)


mlandauer:
Oh what the hell. I'll go ahead and merge this and run it on the server

henare:
@mlandauer sorry I didn't see this earlier, I was AFK. You can always ping me in Slack if there's something time-sensitive :clock10:

PR #1036, opened 2016-07-23
Improve log line capture

mlandauer:
This significantly simplifies the way that log lines are captured from the running scrapers. It also has the advantage of making the tests more robust (because the line handling is much more consistent now).

Best of all it uses a new feature of more recent versions of docker that allow log line capture to be resumed mid-way rather than having to start from the beginning every time.


mlandauer:
This change requires mysql 5.6.4 or later because it depends on fractional second support for datetime in the database

mlandauer:
Bugger, the server is still on 5.5.49

mlandauer:
Because of https://github.com/openaustralia/morph/pull/1037 the server is now at mysql 5.6.31. So, at least that blocker is now removed.

henare:
@mlandauer Sounds great! I've tested this branch and all seems to be working so I'll reassign to you to merge and deploy.

PR #1039, opened 2016-07-27
Limit log lines to 10,000 lines

### LINKED ISSUE #919: Consider limiting the number of lines of output your scraper can write

mlandauer:
This PR also adds some documentation and amazingly more tests which also makes sure that the counting of log lines can handle restarts of the logging process.

This fixes #919


henare:
Shiny :sparkles: I can't see any obvious problems or improvements.

I'll reassign to you for merge and deploy. Nice one on coming up with this simpler solution :+1:

PR #1044, opened 2016-07-30
Fix referential integrity part 1

mlandauer:
When destroying a scraper will now destroy the associated bits and pieces as well. This PR also adds a new rake task to clean up any records associated with deleted scrapers.

There is a second associated PR which adds foreign key constraints to the database once everything is in a better state.

But before touching that we should get the changes here deployed and the rake task run on the production server

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -71,6 +71,7 @@ gem "stripe"
 gem 'haml-coderay'
 gem 'd3-rails'
 gem "validate_url"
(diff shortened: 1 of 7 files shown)


henare:
LGTM, reassigning to you to merge and deploy.

mlandauer:
Deployed and now running the rake task on the server

PR #1045, opened 2016-08-01
Remove need for intermediate images

mlandauer:
See comment on commit for a little more information...

### DIFF
--- a/lib/morph/docker_runner.rb
+++ b/lib/morph/docker_runner.rb
@@ -65,7 +65,18 @@ module Morph
         copy_config_to_directory(repo_path, dest, false)
         yield(:internalout, "Injecting scraper and running...\n")
 
-        c.archive_in(Dir.glob(File.join(dest, "*")), "/app")
+        # Rather than using archive_in we're doing this more roundabout way
+        # because archive_in seems to have very broken handling of directories
+        # TODO Submit a fix to the docker-api gem to fix this
+        # In the meantime get something more long-winded working here
+
+        tar_file = Docker::Util.create_dir_tar(dest).path
+        File.open(tar_file, "rb") do |tar|
+          c.archive_in_stream("/app") do
+            tar.read(Excon.defaults[:chunk_size]).to_s
+          end
+        end
+        File.unlink(tar_file)
       end
 
       c.start
--- a/lib/morph/docker_runner.rb
+++ b/lib/morph/docker_runner.rb
@@ -189,7 +189,6 @@ module Morph
     # image and return a new image
     def self.inject_files(image, dest)
       Dir.mktmpdir('morph') do |dir|
-        FileUtils.mkdir(File.join(dir, 'app'))
         Morph::DockerUtils.copy_directory_contents(dest, File.join(dir, 'app'))
         docker_build_command(image, ['ADD app /app'], dir) do
           # Note that we're not sending the output of this to the console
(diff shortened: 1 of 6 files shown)


lib/morph/docker_runner.rb
```diff
@@ -41,21 +41,14 @@ def self.compile_and_start_run(
       # If something went wrong during the compile and it couldn't finish
       return [nil, nil] if i3.nil?
 
-      # Insert the actual code (and database) into the container
-      i4 = Dir.mktmpdir('morph') do |dest|
-        copy_config_to_directory(repo_path, dest, false)
-        yield(:internalout, "Injecting scraper and running...\n")
-        inject_files2(i3, dest)
```
henare, line 48: I don't think the `.inject_files2` method is used after this change so it can be removed.

lib/morph/docker_runner.rb
```diff
@@ -67,6 +60,14 @@ def self.compile_and_start_run(
       }
 
       c = Docker::Container.create(container_options)
+
+      Dir.mktmpdir('morph') do |dest|
+        copy_config_to_directory(repo_path, dest, false)
+        yield(:internalout, "Injecting scraper and running...\n")
+
+        c.archive_in(Dir.glob(File.join(dest, "*")), "/app")
```
henare: Will this also copy subdirectories? I ask because some scraper repositories definitely contain subdirectories.

lib/morph/docker_runner.rb
mlandauer, line 48: Ooh yes, true true. I'll make that change and add it to this PR

lib/morph/docker_runner.rb
mlandauer: It should do though I'm not absolutely positive it's tested

mlandauer:
Sure enough there was a bug in the docker api gem with the `archive_in` method with its handling of subdirectories. Thanks @henare for spotting that potential (and real) problem and avoiding mishap.

I've worked around it with a little bit of extra code to keep things moving. Probably worth @henare looking over it again now

lib/morph/docker_runner.rb
```diff
@@ -67,6 +60,25 @@ def self.compile_and_start_run(
...
+      Dir.mktmpdir('morph') do |dest|
+        copy_config_to_directory(repo_path, dest, false)
+        yield(:internalout, "Injecting scraper and running...\n")
+
+        # Rather than using archive_in we're doing this more roundabout way
+        # because archive_in seems to have very broken handling of directories
+        # TODO Submit a fix to the docker-api gem to fix this
+        # In the meantime get something more long-winded working here
+
+        tar_file = Docker::Util.create_dir_tar(dest).path
```
henare: Extract into a method?

lib/morph/docker_runner.rb
```diff
@@ -67,6 +60,25 @@ def self.compile_and_start_run(
...
+        File.open(tar_file, "rb") do |tar|
+          c.archive_in_stream("/app") do
+            tar.read(Excon.defaults[:chunk_size]).to_s
+          end
+        end
+        File.unlink(tar_file)
```
henare: How about changing this method to `.delete`? I had to look "unlink" up because I wasn't sure what it did whereas I definitely would have known what "delete" does.

henare:
I've added a couple more small suggestions. However I've tested this so aside from those it's ready to go. Merge and deploy at your leisure :+1:

PR #1046, opened 2016-08-01
Fix referential integrity part 2

mlandauer:
This adds foreign key constraints. This is the second PR related to #1044. It requires a few of the tests to be adjusted which otherwise were breaking the constraints


henare:
When I ran this locally the migrations failed but I think that's because I hadn't run the rake task to fix existing referential integrity problems. It worked fine on my test database and worked fine when I reset my development database so I think this is good to go.

PR #1049, opened 2016-08-03
Mitmproxy fixes

mlandauer:
This should hopefully now properly update the certificate on the mitmproxy docker image. The source of confusion was that there were two different mitmproxy directories for some inexplicable reason.

I've cleaned that up. So, now there is only one mitmproxy directory. Also, OS X homebrew now has a much more recent version of mitmproxy. Made a small tweak to the script to ensure that it runs under the version on the server as well as the latest os x one.


henare:
Nicely spotted :tada: I would never have found that!

I've tested this and it's working locally. Assigning back to you for merge and deploy.

PR #1054, opened 2016-08-07
Only redirect traffic from morph docker containers to mitmproxy

### LINKED ISSUE #523: mimtproxy causes discourse "launcher bootstrap app" to fail
because of a certificate.

Workaround is to temporarily disable the forwarding to mitmproxy by running `iptables-morph-remove` and then running `iptables-morph-add` when finished.

mlandauer:
This creates a new docker network which is used to run scrapers in. This allows us more fine-grained control over the networking. Best of all we can create a special interface on the host machine 'morph' from which all traffic from scrapers will originate. Then, we can just redirect traffic from that interface to mitmproxy which fixes #523.

This pull request will redirect traffic from the "old" bridge0 network and the "new" morph network to mitmproxy which means that it can be deployed without breaking anything.

When everything is up and running and deployed then we can remove the iptables stuff to redirect the traffic from bridge0 to mitmproxy which will then have properly fixed #523.

### DIFF
--- a/lib/morph/docker_runner.rb
+++ b/lib/morph/docker_runner.rb
@@ -14,6 +14,9 @@ module Morph
       'package.json'
     ]
     BUILDSTEP_IMAGE = 'openaustralia/buildstep'
+    DOCKER_NETWORK = 'morph'
+    DOCKER_BRIDGE = 'morph'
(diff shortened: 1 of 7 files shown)


henare:
My internet is not reliable enough to test this right now. @mlandauer I don't want to block this so I'll assign it back to you to merge and deploy.

Issue #1066, opened 2016-08-12
Even with big amounts of memory sidekiq process is likely getting killed for OOM


mlandauer:
There are some strong indications that this is happening:
1. Running `bundle exec rake app:emergency:show_queue_run_inconsistencies RAILS_ENV=production` shows that there were some runs that hadn't finished which there weren't any jobs in the queue for. This can happen if a running sidekiq process gets killed. According to https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting#my-sidekiq-process-is-crashing-what-do-i-do "every time the Sidekiq process crashes, any messages being processed are lost. You can avoid this with Sidekiq Pro's reliable fetch feature."
2. Running `dmesg -T | egrep -i 'killed process'` shows that on Fri Aug 12 14:02:43 2016 a ruby process was killed for OOM.
3. New Relic shows memory steadily going up until it suddenly drops at the same as above
![image](https://cloud.githubusercontent.com/assets/2221/17614925/adc60b62-60ad-11e6-9fcf-3af8713a1411.png)

mlandauer:
To diagnose this further we need to catch the process in the act. Need to pay closer attention to the new relic warnings about memory > 90%!

mlandauer:
I'm pretty confident that this is fixed now because New Relic is showing a much more stable memory usage and I haven't noticed jobs disappearing from the queue

PR #1069, opened 2016-08-15
Significantly reduce peak memory usage of sidekiq process

mlandauer:
It turns out we've been reading the contents of the entire sqlite database into memory when copying the data out of a finished scraper in the sidekiq RunWorker process.

This is clearly not a good thing. This PR fixes this and instead reads everything into temporary files which get passed around. At no stage does the data read from the container get read into memory as a single blob.

This should hopefully also go some way (if not all the way) towards fixing #1066. However, we'll only really know when we deploy this to production and see what happens.

### DIFF
--- a/lib/morph/docker_runner.rb
+++ b/lib/morph/docker_runner.rb
@@ -112,6 +112,25 @@ module Morph
     # exact timestamp, just ones after it.
     def self.attach_to_run_and_finish(container, files, since = nil,
                                       max_lines = nil)
+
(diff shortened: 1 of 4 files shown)


henare:
>However, we'll only really know when we deploy this to production and see what happens.

Sounds good. I've tested locally and things are still working. Assigning to @mlandauer for merge and deploy.

mlandauer:
deployed to production 🚀

Issue #1061, opened 2016-08-10
Add extra guards against running too many containers at once


mlandauer:
At the moment we're depending on the queue limits to constrain the number of containers that can run at the same time. This actually isn't terribly good and is fundamentally flawed. Here's why:

It's perfectly possible (and expected by design) for the worker on the queue to time-out because the scraper hasn't written anything out to a log for 5 minutes or so. In this case, the job will go into the retry queue and meanwhile another job will get picked up which likely will start up another container taking it over the limit of running containers (currently supposed to be 20)

So, it would make sense to add an extra check in the RunWorker that checks how many containers associated with runs are currently up and running. If there are more than the limit it should throw an exception so it goes in the retry queue.

PR #1072, opened 2016-08-17
Dynamic concurrent scraper setting

mlandauer:
First step in doing #1061. Allows an admin to change the number of concurrent scrapers on the fly from the admin panel.

### DIFF
--- a/app/admin/dashboard.rb
+++ b/app/admin/dashboard.rb
@@ -27,6 +27,17 @@ ActiveAdmin.register_page "Dashboard" do
           button_to "Go into site-wide read-only mode", toggle_read_only_mode_admin_site_settings_path
         end
       end
+      para do
+        form action: update_maximum_concurrent_scrapers_admin_site_settings_path, method: :post do
+          div do
+            label 'Maximum concurrent scrapers'
+            input name: 'maximum_concurrent_scrapers', value: SiteSetting.maximum_concurrent_scrapers
+          end
+          div class: "buttons" do
+            input type: 'submit', value: 'Update'
+          end
+        end
+      end
       span class: "blank_slate" do
         span I18n.t("active_admin.dashboard_welcome.welcome")
         small I18n.t("active_admin.dashboard_welcome.call_to_action")
--- a/app/admin/dashboard.rb
+++ b/app/admin/dashboard.rb
@@ -28,15 +28,7 @@ ActiveAdmin.register_page "Dashboard" do
         end
       end
       para do
-        form action: update_maximum_concurrent_scrapers_admin_site_settings_path, method: :post do
-          div do
-            label 'Maximum concurrent scrapers'
-            input name: 'maximum_concurrent_scrapers', value: SiteSetting.maximum_concurrent_scrapers
-          end
-          div class: "buttons" do
-            input type: 'submit', value: 'Update'
-          end
-        end
+        render "maximum_concurrent_scrapers_form"
       end
       span class: "blank_slate" do
         span I18n.t("active_admin.dashboard_welcome.welcome")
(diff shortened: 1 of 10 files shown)


spec/models/site_setting_spec.rb
```diff
@@ -25,4 +25,30 @@
...
+    end
+  end
+
+  describe ".update_sidekiq_maximum_concurrent_scrapers!" do
+    it "should set the sidekiq value" do
+      SiteSetting.maximum_concurrent_scrapers = 10
+
+      expect(Sidekiq::Queue['scraper']).to receive(:limit=).with(10)
+      SiteSetting.update_sidekiq_maximum_concurrent_scrapers!
+      #expect(Sidekiq::Queue['scraper'].limit).to eq 10
```
henare: Is this supposed to be commented out?

app/controllers/admin/site_settings_controller.rb
```diff
@@ -1,10 +1,22 @@
 module Admin
   class SiteSettingsController < ApplicationController
+    # Having to hand craft the form for post to update_maximum_concurrent_scrapers
+    # so doesn't include the authenticity_token. Not a big deal because it's only
+    # accessible by admins anyway
+    skip_before_filter :verify_authenticity_token, only: [:update_maximum_concurrent_scrapers]
```
henare: Isn't this vulnerable to cross-site request forgery? i.e. an attacker could get an admin to visit a page elsewhere that could update this setting.

app/admin/dashboard.rb
```diff
@@ -27,6 +27,17 @@
           button_to "Go into site-wide read-only mode", toggle_read_only_mode_admin_site_settings_path
         end
       end
+      para do
+        form action: update_maximum_concurrent_scrapers_admin_site_settings_path, method: :post do
+          div do
+            label 'Maximum concurrent scrapers'
```
henare: Could we add some advice to admins about what the recommended value should be? Whenever I've set this I have to get out a pen and paper to work out what I think it should be.

spec/models/site_setting_spec.rb
```diff
@@ -25,4 +25,29 @@
...
+    end
+  end
+
+  describe ".update_sidekiq_maximum_concurrent_scrapers!" do
+    it "should set the sidekiq value" do
+      SiteSetting.maximum_concurrent_scrapers = 10
+
+      expect(Sidekiq::Queue['scraper']).to receive(:limit=).with(10)
+      SiteSetting.update_sidekiq_maximum_concurrent_scrapers!
+    end
```
mlandauer, line 51: Yes, it is. Line 49 replaces that test. I've pushed another commit to remove the commented out line

app/controllers/admin/site_settings_controller.rb
mlandauer: Ah yes, you're right. Okay, best do this properly then.

app/admin/dashboard.rb
mlandauer: I think it's non-trivial to give a recommended value with the things configured how they are currently. For example, even if you know the system memory you don't necessarily know the memory used by all the non-docker stuff on the machine. So, I think the best thing we can really do for now is to add a little helper text to just remind the admin of the amount of memory that each scraper can potentially use. I've added another commit to that effect

app/controllers/admin/site_settings_controller.rb
mlandauer: Added another commit to fix this properly. Thanks for spotting this!

henare: :ok_hand:

henare:
Nice, LGTM!

PR #1074, opened 2016-08-18
Ensure that we're never running more scrapers than we should

### LINKED ISSUE #1061: Add extra guards against running too many containers at once

mlandauer:
This PR fixes #1061 by adding an extra guard in the RunWorker which checks if we have any available slots for starting a new run. If we're restarting an already existing run then it will always allow that through.

### DIFF
--- a/app/models/site_setting.rb
+++ b/app/models/site_setting.rb
@@ -15,7 +15,7 @@ class SiteSetting < ActiveRecord::Base
   end
 
   def self.maximum_concurrent_scrapers
-    read_setting('maximum_concurrent_scrapers')
+    read_setting('maximum_concurrent_scrapers').to_i
   end
 
   def self.maximum_concurrent_scrapers=(n)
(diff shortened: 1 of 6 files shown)


henare:
I've had a quick look at the code and tested scrapers run locally :+1: I'll assign back to you for merge and deploy.

mlandauer:
🚀  Deployed to production

PR #1080, opened 2016-08-24
Simplify connection_logs table to reduce space. Fixes #1079

### LINKED ISSUE #1079: Reduce disk space overhead of storing connection_logs information
Currently the connection_logs table is far bigger (in disk space) by far than any of the other tables. Let's really figure out what the minimum information we need here is really and store that instead so that we can significantly reduce the disk space overhead of running the morph server

mlandauer:
Because it has been taking more than 30 GB in production.

So, removed the following columns:
* response_code
* response_size
* request_size
* path
* scheme
* method
* created_at
* updated_at

@henare would you mind looking at this first thing this morning? I'm really keen on getting this deployed this morning so I can see if we can move back to the smaller server yet which I want to do as soon as humanly possible


henare:
I've reviewed the code. The migration syntax threw me but that's just because I didn't know about The New Way :smile: 

As far as the functional changes goes if you think that's what we need to do then I'm happy to go with that.

A note on the deploy: IIRC by default Capistrano will migrate the DB and then boot the new code. For this change you might want to deploy first and then do the migration. If the old code is running against a migrated DB it will fail whereas the new code should work against and unmigrated DB.

I'll assign back to you for merge and deploy.

PR #1081, opened 2016-08-25
Limit log lines in process

mlandauer:
Much better solution to #919. This limits the log output at the process level by running the scraper within a wrapper script. This ensures that a scraper infinitely producing log lines will not fill up the log file within the docker container.

### DIFF
--- a/lib/morph/docker_runner.rb
+++ b/lib/morph/docker_runner.rb
@@ -70,7 +70,10 @@ module Morph
         }
       ) unless exists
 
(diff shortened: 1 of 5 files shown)


henare:
@mlandauer I'm supposed to be away today and not back until Tuesday so I can't do a proper code review right now. It's almost certainly more important to get this deployed and socialised so I'll assign it back to you. You could also ask @equivalentideas to double-check it and do a quick review if you think it needs it but I'll leave that up to you.

mlandauer:
Okay cool. Thanks @henare. I'll just go ahead and merge and deploy it so that we can see if I've broken anything :-)

PR #1114, opened 2017-01-31
[WIP] Make backups easier to understand. 

auxesis:
**This PR is a Work in Progress. Please don't merge until I say I think it's in an OK state.**

I've refactored the backup script into functions. 

The intention is to make the logic easier to follow, so the operator doesn't need to know all the details of how the backups are performed. 

Currently the only logic change is the number of backups to retain (was previously 5, now set to 1, due to disk space problems per #1104)

### DIFF
--- a/provisioning/roles/backups/files/database-backup.sh
+++ b/provisioning/roles/backups/files/database-backup.sh
@@ -21,11 +21,17 @@ setup_backup() {
   umask 077
 }
 
+# Free up all available disk space before a backup happens.
+free_all_available_disk_space() {
+  docker system prune --all --force
+}
+
 set -e
 BACKUP_DIR="/backups/mysql"
 MAX_BACKUPS=1
 
 setup_backup
+free_all_available_disk_space
 perform_backup
 count_backups
 
--- a/provisioning/roles/backups/files/database-backup.sh
+++ b/provisioning/roles/backups/files/database-backup.sh
@@ -1,19 +1,35 @@
 #!/bin/sh
 
 # Backup the database to /backups/mysql and keep the last five days around
-# This script should be run from a cron job once per days
+# This script should be run from a cron job once a day.
 
+count_backups() {
+  backup_count="$(ls -1 $BACKUP_DIR | wc -l)"
+}
+
+find_oldest_backup() {
+  oldest_backup="$BACKUP_DIR/$(ls -1rt $BACKUP_DIR | head -n 1)"
+}
+
+perform_backup() {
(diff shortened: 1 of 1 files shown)


provisioning/roles/backups/files/database-backup.sh
```diff
@@ -1,19 +1,35 @@
...
-do
-  oldest=`ls -1rt $BACKUP_DIR | head -n 1`
-  rm -rf $BACKUP_DIR/$oldest
-  no_backups=`ls -1 $BACKUP_DIR | wc -l`
+
+setup_backup
+perform_backup
+count_backups
+
+while [ $backup_count -gt 1 ]; do
```
equivalentideas: What's the significance of the change here from `-gt 5` to `-gt 1`?

equivalentideas:
@auxesis I've raised the one query I had, but overall the named functions seem like a useful extraction 👍 I can understand what's going on here :)

provisioning/roles/backups/files/database-backup.sh
auxesis: Great question. 

Right now the script makes sure there's never more than 5 backups. 

The new script will make sure there's never more than 1 backup. 

This is done based on the fact that we're constantly deleting old backups, and there are never more than 2 present at any point. 

I reckon we could refactor the number into a `$max_backups` variable though.

equivalentideas: Nice that makes sense. Would be good to mention in the PR description that it's not just a refactor ;-)

provisioning/roles/backups/files/database-backup.sh
```diff
@@ -1,19 +1,36 @@
...
-  oldest=`ls -1rt $BACKUP_DIR | head -n 1`
-  rm -rf $BACKUP_DIR/$oldest
-  no_backups=`ls -1 $BACKUP_DIR | wc -l`
+MAX_BACKUPS=1
+
+setup_backup
+perform_backup
+count_backups
+
+while [ $backup_count -gt $MAX_BACKUPS ]; do
```
auxesis, line 38: @equivalentideas let me know what you think of naming the variable.

equivalentideas: All seem clear to me. Is there some convention for the all caps/small caps difference? I'm not familiar with shell scripting at all by the way :)

mlandauer:
As there hasn't been any activity on this WIP PR in a while I'm closing this

PR #1118, opened 2017-02-02
Upgrade Python bits so newer docker module works

auxesis:
Ansible's [`docker`](http://docs.ansible.com/ansible/docker_module.html) module has been deprecated in favour of the [`docker_container`](http://docs.ansible.com/ansible/docker_container_module.html) and [`docker_image`](http://docs.ansible.com/ansible/docker_image_module.html) modules. 

The mitmproxy role uses the deprecated `docker` module to set up mitmproxy. Post the Docker nuke (#1104), applying this role fails due to a bunch of Docker IPC differences. 

This PR switches the mitmproxy role to use the newer `docker_container` module. To do this, there's a bunch of pip upgrades that need to happen, which this PR handles.

### DIFF
--- a/provisioning/roles/mitmproxy/tasks/main.yml
+++ b/provisioning/roles/mitmproxy/tasks/main.yml
@@ -1,28 +1,47 @@
-- name: Make sure that pip is installed
+- name: Ensure system pip is installed
   apt: pkg=python-pip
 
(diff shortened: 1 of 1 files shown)


auxesis:
@henare correct, it's already been rolled out on the Morph box. ✅ 

Will merge now.

PR #1128, opened 2017-03-10
Change SQL queries to use double quotes

handelaar:
SQLite prefers double quotes around table and column names. Use of single quotes in select_first_ten() here causes the default example query in API docs to fail if a user edits it to add (eg) an ORDER BY if that user also uses single quotes, as the example encourages.

Then also changed the other three explicit queries here for the sake of constistency

### DIFF
--- a/lib/morph/database.rb
+++ b/lib/morph/database.rb
@@ -128,7 +128,7 @@ module Morph
 
     # Returns 0 if table doesn't exists (or there is some other problem)
     def no_rows(table = table_names.first)
-      q = sql_query_safe("select count(*) from '#{table}'")
+      q = sql_query_safe("select count(*) from \"#{table}\"")
       q ? q.first.values.first : 0
     rescue *CORRUPT_DATABASE_EXCEPTIONS
       0
@@ -163,16 +163,16 @@ module Morph
     end
 
     def standardise_table_name(table_name)
-      sql_query_safe("ALTER TABLE '#{table_name}' " \
-        "RENAME TO '#{Database.sqlite_table_name}'", false)
+      sql_query_safe("ALTER TABLE \"#{table_name}\" " \
+        "RENAME TO \"#{Database.sqlite_table_name}\"", false)
     end
 
     def select_first_ten(table = table_names.first)
-      "select * from '#{table}' limit 10"
+      "select * from \"#{table}\" limit 10"
     end
 
     def select_all(table = table_names.first)
-      "select * from '#{table}'"
+      "select * from \"#{table}\""
     end
 
     def first_ten_rows(table = table_names.first)


lib/morph/database.rb
```diff
@@ -128,7 +128,7 @@ def sql_query_safe(query, readonly = true)
 
     # Returns 0 if table doesn't exists (or there is some other problem)
     def no_rows(table = table_names.first)
-      q = sql_query_safe("select count(*) from '#{table}'")
+      q = sql_query_safe("select count(*) from \"#{table}\"")
```
henare, line 131: Rather than using backslashes to escape the double quote I think this would be more readable if you use % notation. That would make this line:

```ruby
q = sql_query_safe(%{select count(*) from "#{table}"}) 
```

lib/morph/database.rb
```diff
@@ -163,16 +163,16 @@ def write_sqlite_database(content)
...
     def standardise_table_name(table_name)
-      sql_query_safe("ALTER TABLE '#{table_name}' " \
-        "RENAME TO '#{Database.sqlite_table_name}'", false)
+      sql_query_safe("ALTER TABLE \"#{table_name}\" " \
+        "RENAME TO \"#{Database.sqlite_table_name}\"", false)
     end
 
     def select_first_ten(table = table_names.first)
-      "select * from '#{table}' limit 10"
+      "select * from \"#{table}\" limit 10"
```
henare, line 171: While we're making this change what do you think about also using the convention of capitalising the keywords?

lib/morph/database.rb
handelaar: Not enough rubyfu here to apply same to the other affected queries in this file. Obv I have no issue with it

lib/morph/database.rb
handelaar: Let's be clear: I'm proposing this solely because it's a moderately-serious UI fail - where this query is reprinted in the API documentation it *encourages* users to use single quotes, the first thing I tried was to add an ORDER BY to the example, and with single quotes that query outright fails to work.

So yeah, by all means, but this isn't rooted in me being anal about SQL style, but about whether the API works as documented. Right now, it (slightly) does not.

equivalentideas:
Hey @handelaar :) why'd you close this?

handelaar:
You won't take it, and I'm in no position to both change what you want changing *and* be sure it will work. Those things being true, this has become nothing but clutter. :)

equivalentideas:
> You won't take it, and I'm in no position to both change what you want changing and be sure it will work. Those things being true, this has become nothing but clutter. :)

Fair enough :) I've opened this as an issue so hopefully someone does get to it and can use everything here: #1144

Issue #890, opened 2015-08-12
Add more metadata to package.json for node scrapers created via the site


drzax:
At a minimum, the required keys 'name' and 'version' should be added.

henare:
Excuse my ignorance :) What's the reason for this?

drzax:
I'll grant you that there's probably no technical reason this is necessary. It's basically to conform with the specification: https://docs.npmjs.com/files/package.json

henare:
Cool, thanks, that makes sense.

equivalentideas:
This looks like the template file in question https://github.com/openaustralia/morph/blob/master/default_files/nodejs/template/package.json

@simon would you care to drop in here what the values should be?

drzax:
I'm happy to submit a pull request. What data does can I use in the template and what format should the placeholders take?

equivalentideas:
@drzax Sounds like you're the expert on this one ;-) Whatever would be helpful to users, and if there's a common best practice / convention that sounds good.

drzax:
Does the template get run through something which replaces tokens though? e.g. It would Ideally include something like:

```
{
  "name": "${scraper-slug}",
  "description": "${scraper-description}",
  "author": "${github-user} <${github-email}>"
...etc
}
```

But I don't know what data is available in the template.

equivalentideas:
Oh good question. If you can find where it's being inserted there's probably a way to parse variables to the template.

drzax:
I suspect if any templating of default files were to happen it would happen here: https://github.com/openaustralia/morph/blob/master/lib/morph/language.rb#L102

So it doesn't appear that there is any and I don't think that's something I have the Ruby knowledge to change.

benrfairless:
Thanks for raising this, and sorry it has sat without a response for so long.

We are tidying up the morph.io issue tracker. This issue dates from 2015 and, on review, looks to have gone stale (the idea has been overtaken, superseded, or not picked up in the years since), so we are closing it as part of a backlog review to keep the tracker focused on work we can realistically act on. Closing it is not a judgement that the underlying idea was wrong.

If this is still relevant to you, please reopen it or open a fresh issue with up to date detail and we will gladly take another look.

This triage was carried out with AI assistance (Claude Opus 4.8 via Claude Code) and reviewed by the OpenAustralia Foundation team.

PR #1138, opened 2017-03-24
Add sensible defaults to package.json

drzax:
Add a couple more default values to package.json for nodejs scrapers created on the site.

Ideally, this would include quite a bit more detail see: #890

### DIFF
--- a/default_files/nodejs/template/package.json
+++ b/default_files/nodejs/template/package.json
@@ -1,7 +1,13 @@
 {
-	"dependencies": {
-		"cheerio": "latest",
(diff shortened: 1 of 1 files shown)


henare:
@drzax does this close #890?

drzax:
I don't think it should because I think other fields we know values for (like authorship, description and repository location) should be populated. But that involves other work to get a templating system going (see [comments starting here](https://github.com/openaustralia/morph/issues/890#issuecomment-288910935)). 

But I'll let someone closer to the project decide if that's something worth doing.

Issue #875, opened 2015-07-20
Anchor links make title hidden behind the nav


henare:
e.g. if you go to https://morph.io/documentation#format then the "Storing your data" header is hidden behind the nav bar. This is probably some CSS I've screwed up in my adventures :-/

henare:
Fixed in #1139.

PR #1139, opened 2017-03-28
Fix hidden headers behind navbar caused by anchor links in documentation

Anishka0107:
This pull request aims to fix the issue https://github.com/openaustralia/morph/issues/875
When anchor links are used, the header gets hidden behind the navigation bar.
It is caused by default in Bootstrap, and has been corrected in this commit by overriding the default values.
This commit does not change the spacing values of this page or any other pages; only the anchor links are made visible by shifting below.

For eg, if we open the page https://morph.io/documentation#languages

Initially, it looked like

uploadorig
-----------------------------------------------------------------------------------------

After the commit, it would look like,

uploadini

### DIFF
--- a/app/assets/stylesheets/bootstrap_and_overrides.scss
+++ b/app/assets/stylesheets/bootstrap_and_overrides.scss
@@ -49,3 +49,8 @@ body {
 .jumbotron p {
   font-weight: normal;
 }
+
+.section {
+  padding-top: 50px;
+  margin-top: -30px;
+}


henare:
Deployed and confirmed fixed :rocket: 

@Anishka0107 a tip for next time :smile: if you include a line in your pull request description that says:

Fixes #875.

Then that issue will automatically close when your pull request is closed :sparkles:

Issue #983, opened 2016-04-08
Don't show the clear Database settings action if the scraper has no data


equivalentideas:
It's confusing to have a setting that does nothing. There are some important and irreversible actions on the settings page, the simpler we can make it for the users current context the better.

mlandauer:
It would be even better in that circumstance if the button was disabled rather than removed. That way a person learns that the button is there but it's clear that it doesn't do anything in that situation.

PR #1140, opened 2017-03-29
Disable Clear database button if scraper is empty

Anishka0107:
This pull request aims to fix the issue https://github.com/openaustralia/morph/issues/983
If the scraper is empty, the button to clear the database on the settings page must be disabled for ease to the user. 
So, in this commit, the design and action of the button has been changed to make it disabled if the scraper is empty.

### DIFF
--- a/app/views/scrapers/_danger.html.haml
+++ b/app/views/scrapers/_danger.html.haml
@@ -16,9 +16,14 @@
                   of scraped data will be gone!
                 - else
                   All your data will be gone!
-              = button_to clear_scraper_path(@scraper), {class: "btn btn-warning", data: {confirm: "All the data you’ve scraped will be deleted. Are you sure you want to clear your scraper’s database?"}} do
-                Clear database
-                %i.glyphicon.glyphicon-th-list
+              - if @scraper.has_data?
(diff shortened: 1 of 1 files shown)


app/views/scrapers/_danger.html.haml
```diff
@@ -16,9 +16,14 @@
                   of scraped data will be gone!
                 - else
                   All your data will be gone!
-              = button_to clear_scraper_path(@scraper), {class: "btn btn-warning", data: {confirm: "All the data you’ve scraped will be deleted. Are you sure you want to clear your scraper’s database?"}} do
-                Clear database
-                %i.glyphicon.glyphicon-th-list
+              - if @scraper.has_data?
+                = button_to clear_scraper_path(@scraper), {class: "btn btn-warning", data: {confirm: "All the data you’ve scraped will be deleted. Are you sure you want to clear your scraper’s database?"}} do
```
equivalentideas, line 20: You should be able to remove all the repetition of the button html by making the condition only cover the attribute we care about here. Something like `button_to {disabled: ( "disabled" unless @scraper.has_data? )}`

mlandauer:
I'm closing this PR as there hasn't been any activity on it in a while

Issue #1084, opened 2016-09-07
Cannot delete scraper due to foreign key constraint failing


henare:
```
[Morph/production] ActiveRecord::StatementInvalid: Mysql2::Error: Cannot delete or update a parent row: a foreign key constraint fails (`morph`.`metrics`, CONSTRAINT `fk_rails_14230a8cc7` FOREIGN KEY (`run_id`) REFERENCES `runs` (`id`)): DELETE FROM `runs` WHERE `runs`.`id` = 168069
```
## Backtrace

```
line 133 of [PROJECT_ROOT]/app/controllers/scrapers_controller.rb: destroy
```

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/35868/faults/31032607)

tmtmtmtm:
I've also noticed that I can't delete scrapers that have Webhooks defined. Once I delete those, I can then delete the scraper.

equivalentideas:
I think the issue here might be that we're trying to delete the Run(s) associated with the scraper, but there are Metric(s) that reference those runs.

equivalentideas:
I think the issue here might be that we're trying to delete the Run(s) associated with the scraper, but there are Metric(s) that reference those runs.

chrismytton:
I've just been looking at this locally.

It seems that when a run has _multiple_ metrics associated with it, trying to destroy the run causes an error. This is because Run is supposed to `have_one` Metric, but this isn't being enforced at the database level, so there are certain runs that have more than one metric associated with them.

I've opened https://github.com/openaustralia/morph/pull/1145 with a failing test to demonstrate this problem.

equivalentideas:
@chrismytton nice sloothing :)

> certain runs that have more than one metric associated with them

Any idea how a Run gets into this state?

chrismytton:
> Any idea how a Run gets into this state?

Good question! I've just had a quick grep around and it looks like [these lines in `Morph::Runner`](https://github.com/openaustralia/morph/blob/b921061a680b71aafa2d911fa825819561cb8222/lib/morph/runner.rb#L147-L148), which create a metric for a run, are the primary suspect. Working back up the stack leads me to [this line in `RunWorker`](https://github.com/openaustralia/morph/blob/05b847940be48f43a20ca624051bc273e573836c/app/workers/run_worker.rb#L10), which finds a run and then passes it to `Morph::Worker`.

[...]

PR #1145, opened 2017-04-12
Fix error when trying to delete scrapers

chrismytton:
## What does this do?

This fixes the problem with deleting scrapers due to a foreign key error that was originally reported in https://github.com/openaustralia/morph/issues/1084.

## Why was this needed?

When a run has multiple metrics associated with it, trying to destroy it causes an error. This is because Run is supposed to `have_one` Metric, but this isn't being enforced at the database level, so there are certain runs that have more than one metric associated with them.

## Implementation notes

For now, I've fixed this at the application level. I attempted to write a migration to add a database constraint, but it proved tricky because there's already an index on the `metrics.run_id` column and my mysql/rails migration skills fell short of a satisfying solution. 

I'm happy to be persuaded that we should still add a database constraint/script to tidy up old data, but perhaps we can do that as a separate issue with a separate pull request? Either way, I think there's value in getting this live so that we can fix the user-facing error, at least.

## Related issues

Fixes https://github.com/openaustralia/morph/issues/1084

### DIFF
--- a/app/models/run.rb
+++ b/app/models/run.rb
@@ -16,6 +16,8 @@ class Run < ActiveRecord::Base
 
   before_create { |run| run.build_metric }
 
+  before_destroy { |run| Metric.where(run_id: run.id).destroy_all }
+
   # TODO: Run requires an owner - add a validation for that
 
   def database
--- a/app/models/run.rb
+++ b/app/models/run.rb
@@ -14,6 +14,8 @@ class Run < ActiveRecord::Base
            to: :scraper, allow_nil: true
   delegate :utime, :stime, :cpu_time, to: :metric, allow_nil: true
 
+  before_create { |run| run.build_metric }
+
   # TODO: Run requires an owner - add a validation for that
 
   def database
--- a/lib/morph/runner.rb
+++ b/lib/morph/runner.rb
@@ -144,8 +144,7 @@ module Morph
       end
 
       # Now collect and save the metrics
-      metric = Metric.create(result.time_params) if result.time_params
-      metric.update_attributes(run_id: run.id) if metric
+      run.metric.update_attributes(result.time_params) if result.time_params
 
       # Update information about what changed in the database
       diffstat = Morph::SqliteDiff.diffstat_safe(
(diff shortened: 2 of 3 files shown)


henare:
Nice one @chrismytton :fire: As you say on #1084, I think the way to go is:

>So I think we probably want to do some kind of find_or_create type dance in Morph::Runner to check if a metric already exists for the run, and if it does then just update that rather than creating a new one?

That will prevent multiple Metrics being created at the application-level. Then we should also introduce a database constraint. Finally we should also write a migration to clean up any problems in the existing data :shower:

chrismytton:
I've updated the pull request description to explain what I've done. So I think this is ready for review now 🔍

henare:
@chrismytton thanks for all of your clear explanations in the pull request body and in each commit message. That's made this really easy to review. Let's give this a shot :+1: I'll merge and deploy :rocket:

henare:
This is deployed.

After I deployed it the current running scrapers started to throw an exception:

```
NoMethodError: undefined method `update_attributes' for nil:NilClass

    [PROJECT_ROOT]/lib/morph/runner.rb:147 :in `attach_to_run_and_finish` 
```

This is because the existing Runs didn't have Metrics already associated with them. I just did the hacky thing and ran this on the console to create the Metrics:

    Scraper.running.each { |s| s.last_run.build_metric }

And things seem to be running OK. I've had it recur a few times and I'm not sure what's up with that. I just manually intervened again. I'll keep an eye on it for the next few hours.

@chrismytton can you please test deleting a problematic scraper and report back if this fix has worked?

tmtmtmtm:
I successfully deleted four scrapers that had previously been failing, so it looks like this is working.

However, it's *very* slow (I didn't time it exactly, but I'd say it took at least a minute or two). I haven't looked at the code at all, but my initial guess would be a missing index on a reasonably big table.

chrismytton:
> @chrismytton can you please test deleting a problematic scraper and report back if this fix has worked?

Yep, same as @tmtmtmtm, a tad slow, but all working fine now, yay! 🎉

Issue #1135, opened 2017-03-21
Scrapers can't connect to the internet with mitmproxy enabled


henare:
Scrapers can't connect to the internet when mitmproxy is enabled. I think it's because the newish dedicated [morph network](https://github.com/openaustralia/morph/blob/581dffb2e1188457c9933892cb4a51a592b83c20/lib/morph/docker_runner.rb#L19) (192.168.0.0/16) conflicts with the virtualbox bridge network (192.168.11.0/24).

henare:
Sigh. I'm seeing it on production now too. I'm guessing it's the updates we did earlier.

henare:
Here's the docker upgrade:

    docker-engine:amd64 (1.13.0-0~ubuntu-precise, 17.03.0~ce-0~ubuntu-precise)

and the full upgrade log for debugging:

```
[...]

PR #1148, opened 2017-05-09
Comment out statistic on web pages scraped on landing page 

mlandauer:
Because the collection of the statistic is currently broken :-(


henare:
@mlandauer thanks for this PR and thanks for waiting so long for it to be reviewed :hourglass_flowing_sand: 

I definitely agree with the intent - the zero figure is glaring, wrong, and off-putting to people wanting to use morph.io. It would have been nice to conditionally show each stat if we had data for it but that looks hard to do because of the bootstrap column classes.

I'll merge this and make a note on #1135 to revert this once it's fixed.

PR #1156, opened 2017-08-28
Update encrypted file handling

auxesis:
Update the documentation on how to handle encrypted files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,8 +136,8 @@ git clone -n https://github.com/openaustralia/morph
 cd morph
 
 # set up gitcrypt
-git config gitcrypt.salt 'wm2jtxCAfNmv'
-git config gitcrypt.pass '7idce*4e,XDkB/n8,DLLdngMA=6pfE/WvzFdZ67f'
+git config gitcrypt.salt 'YOUR_SALT'
+git config gitcrypt.pass 'YOUR_STRONG_PASSKEY'
 git config gitcrypt.cipher aes-256-ecb
 git config filter.encrypt.smudge "gitcrypt smudge"
 git config filter.encrypt.clean "gitcrypt clean"
--- a/README.md
+++ b/README.md
@@ -144,7 +144,9 @@ git config filter.encrypt.clean "gitcrypt clean"
 git config diff.encrypt.textconv "gitcrypt diff"
 ```
 
-For the real production `salt` and `pass` values, check with the folks in #morph on the [OpenAustralia Slack](https://oaf.slack.com).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -122,7 +122,29 @@ This section will not be relevant to most people. It will however be relevant if
...
+# set up gitcrypt
+git config gitcrypt.salt 'wm2jtxCAfNmv'
+git config gitcrypt.pass '7idce*4e,XDkB/n8,DLLdngMA=6pfE/WvzFdZ67f'
+git config gitcrypt.cipher aes-256-ecb
+git config filter.encrypt.smudge "gitcrypt smudge"
+git config filter.encrypt.clean "gitcrypt clean"
+git config diff.encrypt.textconv "gitcrypt diff"
+```
+
+For the real production `salt` and `pass` values, check with the folks in #morph on the [OpenAustralia Slack](https://oaf.slack.com).
```
equivalentideas: @auxesis I recon this could be clearer with two changes:

1. Move this up into the code block above these lines as a comment. This is important info so should be as close as possible to the bit it relates to, especially for new contributors who might jump ahead as they're reading.
2. I recon we should change the examples to be more clearly examples and not values you might use. e.g. `'YOUR_SALT', 'YOUR_STRONG_PASSKEY`. Maybe we should even give an example of how to generate them? Maybe that's over kill for now :)

auxesis: Makes sense! I'll make those changes now.

auxesis:
@equivalentideas I think this whole section of documentation is a bit of a mess. 

> by ignoring the encrypted files, how will people have the files to be decrypted? That's not clear in these instructions.

It won't work unless you have all the credentials in place, and the only credentials we ship are encrypted ones. There's no documentation about how to set up credentials from scratch when setting up a local dev environment. 

> Do you think we should move this bit up in the readme so people see it before they clone? Or is it really just for us? Maybe we should start a new section, 'Setting up to deploy to the production instance at https://morph.io'?

This section of documentation should be relevant to to any user who wants to set up a public-facing copy of Morph, either for production use or testing. 

Right now, the documentation only helps us. 

It comes back to what I said above: we need to write documentation that covers: 

 - How to set up a production-like environment
 - How to manage credentials

auxesis:
@equivalentideas and to be clear: I think that new documentation should be handled in a separate PR.

equivalentideas:
That all sounds very sensible to me @auxesis , I'll merge this.

henare:
Relates to #948.

PR #1185, opened 2018-04-10
WIP: JP tweaks and twiddles

jamezpolley:


provisioning/playbook.yml
```diff
@@ -29,9 +40,43 @@
         update_cache: yes
         cache_valid_time: 300
         upgrade: yes
+    # Create fake let's encrypt directories when in development
```
mlandauer: Sorry I implemented exactly the same thing on master so this shouldn't be necessary anymore. See https://github.com/openaustralia/morph/commit/6e59591dc6a196abb2e11cbec4cb9642da35bad7

mlandauer:
@james i'm going to close this PR without merging it as I think the changes are being split out into other PRs.

jamezpolley:
Yep, thanks. This was up here mostly for tracking so that the things I was playing with weren't just local.

PR #1207, opened 2019-02-25
Enable multiple platforms

jamezpolley:
The goal of this change is to allow a scraper to pick from a set of
supported platforms by use of the `platform` file. If no platform is
selected, `openaustralia/buildstep:latest` is implicity used, matching
current behaviour. I've already created
https://github.com/openaustralia/buildstep/tree/early_release as an
alternate platform. Although this is not yet on Dockerhub, manual
testing has confirmed that `openaustralia/buildstep:early_release`
does get used if the image already exists locally.

Other potential platform tags might be `heroku_16` and `heroku_18`; or
`herokuish_v0.4.8`; or `deprecated` for people who still need
`cedar:14` after we stop supporting that.

### DIFF
--- a/app/models/scraper.rb
+++ b/app/models/scraper.rb
@@ -266,6 +266,16 @@ class Scraper < ActiveRecord::Base
     Morph::Database.new(data_path)
   end
 
+  def platform
+    platform_file = "#{repo_path}/platform"
+    if File.exist?(platform_file)
+      platform = File.read(platform_file)
+    else
+      platform = "latest"
+    end
+    platform.chomp
+  end
+
   # It seems silly implementing this
   def self.directory_size(directory)
     r = 0
--- a/lib/morph/docker_runner.rb
+++ b/lib/morph/docker_runner.rb
@@ -34,8 +34,14 @@ module Morph
     end
 
     def self.compile_and_start_run(
-      repo_path, env_variables, container_labels, max_lines = 0, platform = "latest"
-    ) 
+      repo_path, env_variables, container_labels, max_lines = 0, scraper = nil
+        )
+      if scraper.nil? || scraper.platform.nil?
+        platform = "latest"
+      else
+        platform = scraper.platform
+      end
+
       i = buildstep_image(platform) do |c|
         yield(:internalout, c)
       end
--- a/lib/morph/docker_runner.rb
+++ b/lib/morph/docker_runner.rb
@@ -27,14 +27,16 @@ module Morph
       512 * 1024 * 1024
     end
 
-    def self.buildstep_image
-      Morph::DockerUtils.get_or_pull_image(BUILDSTEP_IMAGE)
+    def self.buildstep_image(platform = "latest")
+      image = "#{BUILDSTEP_IMAGE}:#{platform}"
+      puts "Using image #{image}"
+      Morph::DockerUtils.get_or_pull_image(image)
     end
 
     def self.compile_and_start_run(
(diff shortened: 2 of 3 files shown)


jamezpolley:
... need to write some tests ..

mlandauer:
@jamezpolley what's the reasoning for having the "platform" configuration as a file in the scraper rather than in the ui of morph as a setting on the scraper?

So far all the stuff in the scraper repositories is not specific to morph. You can run it completely independently and all the files that tell it what version of the language to use, etc.. are all the standard way you do this for every language.

So, to me it seems a little different to have a new file "platform" which is morph specific being something that you put in your code.

In the scraper settings in the UI you can currently change:
* Whether the scraper is auto-run
* Any secret environment variables
* Any webhooks

These are quite morph specific settings and a platform setting there would seem like a fairly natural fit. 

You could also explain in the GUI what the different values are and what the defaults are without having to write more esoteric documentation (that likely almost nobody will read)

jamezpolley:
> So far all the stuff in the scraper repositories is not specific to morph.

In my head, the platform isn't specific to morph either; choosing cedar:14 or heroku:16 is analogous to choosing python 2.7 vs python 3.6

But `early_release` is clearly a morph-specific value.... and in reality the Morph platform, although built on cedar:14 or heroku:16, is always going to be something extra on top. Hrm.

However, it still feels to me like it fits with the scraper rather than in morph, because in many cases this value will be necessary for the scraper to work at all. At the moment, I can clone just about any scraper into my own account, hit run on morph, and it will work out of the box - the only exception being if I need to provide a secret that's used to log into a site, which can't be distributed as part of the repo because it's specific to me. Having certain scrapers that will only work after you clone and then go and set a value in the UI doesn't feel right to me - any more than it would feel right to have to clone a scraper repo and then manually tell Morph that it's python-3.6

> In the scraper settings in the UI you can currently change:
> * Whether the scraper is auto-run
> * Any secret environment variables
> * Any webhooks

I think those things are all different, because they're things that will change for each user who clones the same scraper. If I write a scraper that logs into github and grabs new notifications, that's going to need a secret for an auth token; I'm going to want it auto-run, and I might want it to notify my personal pushbullet account every time it runs. If someone clones the repo from me, they might not want it to auto-run; I definitely don't want them to have my auth secret, and I don't want them to hit my webhook. I *do* want them to know what version of Python they need, any extra packages they need to install, and which flavor of docker image is required to run the scraper.

----

Maybe "platform" is the wrong noun to use here; perhaps the right thing to do would be make this "docker_image"; so that in our case we'd use "openaustralia/buildstep:early_release" rather than just "early_release". OpenAustralia's morph instance could require that any image specified match  "openaustralia/buildstep:.*", other morph instances could have a different requirement. People wanting to run the scraper locally should be able to use this file to grab the known-working docker image and run it locally.

Would that make more sense?

tmtmtmtm:
More as a passing comment, but the point that scrapers are essentially independent of Morph isn't quite true. The requirement that all environment variables must start `MORPH_` means that any scrapers that examine those need to be "morph-aware", and either check for multiple different variables, or be supplied with a morph-specific variable even when running outside of the morph environment.

jamezpolley:
I'm still a bit ambivalent about this. I definitely don't think my current implementation is perfect, but it seems to be good enough for now, and it means I can move ahead with some of the other issues we have at the moment - like the fact that the current buildstep environment is about to hit EOL, and the fact that PHP is entirely broken.

I'm going to merge this now and revisit this once the other issues are fixed.

PR #1208, opened 2019-02-25
Simpler local deploy - push directly from local repo

jamezpolley:
I've found that this enables me to iterate much faster, as I can push
something and try it out without needing to round-trip through Github.

### DIFF
--- a/config/deploy/local.rb
+++ b/config/deploy/local.rb
@@ -1,6 +1,20 @@
 set :stage, :local
 set :rails_env, :production
 
+before :deploy, :deploy_from_local_repo
+
+task :deploy_from_local_repo do
+  set :repo_url,  "file:///tmp/.git"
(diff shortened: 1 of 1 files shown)


mlandauer:
I had a vague memory of capistrano having built-in support for deploying local files, rather than going through a checkout from git, but I don't see it when I just had a quick look. Maybe it was something in capistrano 2?

Anyway, it doesn't really matter. Your change looks good to me and certainly seems like it would be super useful. Thanks!

jamezpolley:
Yes, Capistrano did used to support this, but it was removed because the maintainer didn't use it personally and didn't care to maintain it any more. https://github.com/capistrano/capistrano/issues/695#issuecomment-26487649

This workaround is one that I've seen variants of in a few places and seems to be fairly standard. There's a capistrano plugin (https://github.com/xuwupeng2000/capistrano-scm-gitcopy) or three which do similar things which might be preferable.

PR #1209, opened 2019-02-25
Update dependencies and packages

jamezpolley:
Various tweaks and updates I've had to do to get my local dev instance up and running.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ file at `~/.infrastructure_ansible_vault_pass.txt`. This is the same password as
 
 Install [Vagrant](http://www.vagrantup.com/), [VirtualBox](https://www.virtualbox.org) and [Ansible](http://www.ansible.com/).
 
-Install the hostsupdater plugin: `vagrant plugin install vagrant-hostsupdater`
+Install a couple of Vagrant plugins: `vagrant plugin install vagrant-hostsupdater vagrant-disksize`
 
 Run `vagrant up local`. This will build and provision a box that looks and acts like production at `dev.morph.io`.
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 Ruby 2.3.1, Docker, MySQL, SQLite 3, Redis, mitmproxy.
 (See below for more details about installing Docker)
 
-Development is supported on Linux and Mac OS X.
+Development is supported on Linux (Ubuntu 16.04 works best; Ubuntu 18.04 is possible with some setup) and Mac OS X.
 
 ### Repositories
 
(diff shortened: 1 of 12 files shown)


.ruby-version
```diff
@@ -1 +1 @@
-ruby-2.3.1
+2.3.1
```
jamezpolley: I'm not sure if this is something that should be pushed; it was done for me to get rbenv working locally but I think we use rvm on the server so this should probably remain as it is.

README.md
```diff
@@ -15,7 +15,7 @@
 Ruby 2.3.1, Docker, MySQL, SQLite 3, Redis, mitmproxy.
 (See below for more details about installing Docker)
 
-Development is supported on Linux and Mac OS X.
+Development is supported on Linux (Ubuntu 16.04 works best; Ubuntu 18.04 is possible with some setup) and Mac OS X.
```
jamezpolley, line 18: I think the changes I've made take care of this, so this comment can probably be reverted.

.ruby-version
mlandauer: If you're not sure and you haven't tested it then revert it.

README.md
mlandauer: Then revert it ;-)

README.md
```diff
@@ -128,12 +128,26 @@ file at `~/.infrastructure_ansible_vault_pass.txt`. This is the same password as
 
 #### Production devops development
 
+> This method defaults to creating a 4Gb VirtualBox VM, which can strain an 8Gb Mac. We suggest tweaking the Vagrantfile to restrict ram usage to 2Gb at first, or using a machine with at least 12Gb ram.
+
 Install [Vagrant](http://www.vagrantup.com/), [VirtualBox](https://www.virtualbox.org) and [Ansible](http://www.ansible.com/).
 
-Install the hostsupdater plugin: `vagrant plugin install vagrant-hostsupdater`
+Install a couple of useful Vagrant plugins: `vagrant plugin install vagrant-hostsupdater vagrant-disksize`
```
mlandauer: If the plugin vagrant-disksize isn't an absolute requirement but a nice to have then explain what it does and why it's useful

README.md
```diff
@@ -128,12 +128,26 @@ file at `~/.infrastructure_ansible_vault_pass.txt`. This is the same password as
...
+
+If on Ubuntu, install libreadline-dev: `sudo apt install libreadline-dev libsqlite3-dev`
+
+Install the required ruby version: `rbenv install`
+
+Install capistrano: `gem install capistrano`
 
 Run `vagrant up local`. This will build and provision a box that looks and acts like production at `dev.morph.io`.
 
+_The first time you're doing this, check `provisioning/requirements.yml` for instructions on installing modules from ansible galaxy_
```
mlandauer, line 149: Really nice addition. The information just where you need it :-)

provisioning/playbook.yml
```diff
@@ -21,6 +21,17 @@
   - name: install python 2
     raw: test -e /usr/bin/python || (apt -y update && apt install -y python-minimal)
 
+# Some features in the playbook need ansible2.5, so let's check for that first
+
+
+- hosts: all
```
mlandauer, line 27: Nice! Making things more foolproof

README.md
mlandauer: Ah I see it is a new requirement to make the disk bigger automatically. Maybe get rid of the word "useful" in the sentence "Install a couple of useful Vagrant plugins". It makes it sound like they're not strictly necessary but I think that's not the case. They're kind of both essential, right?

README.md
jamezpolley: Checking again - still required, so I'll leave it.

PR #1211, opened 2019-03-19
First draft of a pain-points overview

jamezpolley:


rbtcollins:
So I'm not sure I'd call this a points of pain overview so much as good docs ;P.

We've discussed a bunch offline; two thoughts that are really only relevant here: your scheduler failure modes are simple bugs; they should be fixed in-situ I think, because that can be done quickly (to whit: don't de-and-requeue things when there is no work slot available - thats not the task failing; use system metrics to inform work slot availability (e.g. if there is io overload, don't schedule more work); immediately place work when slots are freed up (e.g. schedule work immediately at the end of your cleanup of a work slot), cap exponential backoff (e.g. at 5 minutes), discard work after (say) 10 attempts, and finally implement a quick-reset mechanism to zero the queue and allow an immediate restoration of service without mucking around.

pain-points.md
```diff
@@ -0,0 +1,147 @@
...
+
+Currently, all of Morph runs on a single Linode 32Gb VM. Obviously
+this is a single point of failure; but it also means that the number
+of concurrent scrapers is limited by the available RAM. Each scraper
+container is capped at 512Mb and we have a limit of 30 concurrent
+jobs.
+
+## Memory contention + exponential backoff = every-growing retry queue
+
+Until recently, this VM only had 224Gb RAM. It would relatively often
```
jamezpolley, line 107: s/224/24/

pain-points.md
```diff
@@ -0,0 +1,147 @@
...
+## Logging overhead
+
+Currently we have to schedule and run a parallel job for every docker
+container to scrape its console output and feed it to the DB and
+Faye. These jobs time out, so we have to have another job which looks
+for running containers without a log-scraper and start a new
+log-scraping job. This could probably be avoided if we could do
+something like feed the logs into syslog or some other central logging
+system.
+
```
jamezpolley, line 140: ## mitmproxy troubleshooting

Because mitmproxy intercepts HTTPS traffic, we've got a self-signed CA cert on it which needs to be trusted by the scrapers. Most of the customisation on our buildstep image is trying to get this cert in all the places and trusted by all the things.

However, the dev vagrant image doesn't (unless my memory is playing tricks on me) set this up at all; which makes it hard to reproduce problems. Having a dev setup that did set up SSL by default - in as close as possible to the prod setup - would make it easier to troubleshoot problems.

Stepping back a little though: we really only have mitmproxy in place so that we can record the hosts the scraper is scraping from. If we had a different way to record those hosts, we might not need mitmproxy at all.

## Scraper platform updates

At present all scrapers use `openaustralia/buildstep:latest`. This image is based on `cedar:14`, which is old, doesn't work at all with PHP, and about to be EOL. We need to update to `heroku:16`, but this is a breaking change: different set of system packages installed, different versions of languages avalable and so on. [#1207](https://github.com/openaustralia/morph/pull/1207) is the start of a plan to allow scrapers to pick a platform so that we can start with a soft migration to heroku:16, and later make other platforms available.

benrfairless:
Thanks for this early draft, Jamez. We are tidying up the morph.io pull request queue and closing this pain-points overview as it has sat as a draft since 2019 without further work. Closing is not a comment on the value of the idea, just that the draft has gone stale.

If you would like to revive it, please reopen and we will pick the thread back up.

This triage was carried out with AI assistance (Claude Opus 4.8 via Claude Code) and reviewed by the OpenAustralia Foundation team.

PR #1347, opened 2025-08-08
Update SSH authorized users to match main Infra repo

jamezpolley:
I hadn't realised Morph had its own Ansible setup. List of users copied from https://github.com/openaustralia/infrastructure/blob/ea4f88436e1d33c3c3009a864ed7a13fb5216e11/group_vars/all.yml#L26C1-L26C77


simonzippy (approved):
Good to go

PR #1364, opened 2025-10-01
Add ianheggie-oaf to github_users

ianheggie-oaf:
Added ianheggie-oaf to github_users and formatted and sorted list


ianheggie-oaf:
Ended up being duplicated, aside from change to yml style list

PR #1365, opened 2025-10-03
Fix mismatched quotes in github_users list

ianheggie-oaf:
Fix an opsy in last minute change with only one not both double quotes changed to single quote.


ianheggie-oaf:
@Br3nda - I have merged this to fix a syntax error so we can get ssh perms working

Issue #1350, opened 2025-09-29
Setup local Development Environment for morph


ianheggie-oaf:
Setup local Development Environment for morph using docker compose to mimic production environment

related:
* https://github.com/openaustralia/morph/issues/1158

PR #1366, opened 2025-10-09
Get examples and specs working on both old production (16.04)  and new (24.04) docker versions for dev and local staging

ianheggie-oaf:
This PR is the base for me to work on updating ruby on rails and related services. See 

* https://github.com/openaustralia/morph/issues/1350

This PR includes all my work to get the development and specs working on a modern dev system, includes
* updates example to actually run on morph and update the database (adds the h1 headings from example.com) so it is not just testing that it looks like it ran, but that it did actually run on a https site)
* ansible provisioning of vagrant to validate the ansible playbook
* updating docker services to match production versions
* revised and expanded Makefile
* fixed various race and retry issues in ansible
* moved project hard coded values into group_vars so they are easy and obvious to find
* Added various developer aids, eg ENV variables to filter what ansible rules are run, ENV to trigger debug of Docker build

Test results, including pending specs:
```
$ bin/rake
/home/ianh/.local/share/mise/installs/ruby/2.7.6/bin/ruby -I/home/ianh/.local/share/mise/installs/ruby/2.7.6/lib/ruby/gems/2.7.0/gems/rspec-core-3.11.0/lib:/home/ianh/.local/share/mise/installs/ruby/2.7.6/lib/ruby/gems/2.7.0/gems/rspec-support-3.11.0/lib /home/ianh/.local/share/mise/installs/ruby/2.7.6/lib/ruby/gems/2.7.0/gems/rspec-core-3.11.0/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb
Run options: exclude {:slow=>true}

Randomized with seed 49275
....................................*..........*.............................................................................................*.....................................................................................................................................................................................................................................................................................

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) Morph::DockerRunner.compile_and_run is able to run perl example
     # FIXME: Fix perl example test - it works in production but not in test
     Failure/Error: expect(c).not_to be_nil
     
       expected: not nil
            got: nil
     # ./spec/lib/morph/docker_runner_spec.rb:80:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:93:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:91:in `block (2 levels) in <top (required)>'

  2) Morph::DockerRunner.compile_and_run is able to run python example
     # FIXME: Fix python example test - requires heroku-24 platform to be implemented
     Failure/Error: expect(c).not_to be_nil
     
       expected: not nil
            got: nil
     # ./spec/lib/morph/docker_runner_spec.rb:111:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:93:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:91:in `block (2 levels) in <top (required)>'

  3) ApiController#run_remote should test streaming
     # No reason given
     # ./spec/controllers/api_controller_spec.rb:81


Finished in 1 minute 31.13 seconds (files took 2.53 seconds to load)
419 examples, 0 failures, 3 pending

Randomized with seed 49275

JSON Coverage report generated for RSpec to /home/ianh/Projects/OpenAustralia/my_repos/morph/coverage.         1968 / 3645 LOC (53.99%) covered.
```


Dockerfile
```diff
@@ -28,11 +28,6 @@ USER deploy
 COPY --chown=deploy:deploy Gemfile /app/Gemfile
 COPY --chown=deploy:deploy Gemfile.lock /app/Gemfile.lock
 
-# TODO: Update bundler by running "gem install bundler"
```
jamezpolley, line 31: Removing the other lines looks like it's trivially obvious as something that should be done.

Does removing this comment mean that we no longer want/need to do this, or is it just being removed because it's beside the other lines?

Gemfile
```diff
@@ -84,7 +84,7 @@ gem "render_sync", git: "https://github.com/openaustralia/render_sync.git"
 
 # For searchkick 5 we need to install the elasticsearch gem ourselves
 # We're using elasticsearch 7 in production so sticking with the same version for the client
-gem "elasticsearch", "~> 7"
+gem "elasticsearch", "~> 7.17"
```
jamezpolley, line 87: My understanding is that  "~> 7" would allow any minor version, "~> 7.17" is going to only allow sub-versions of 7.17 and not allow 7.18 or higher. Is there a reason we need this?

Makefile
```diff
@@ -1,13 +1,17 @@
-.phony: ansible venv roles help
+.PHONY: all clean deploy help lint \
+        production-deploy production-provision provision \
+        roles services-down services-up \
+        share-web test up vagrant-plugins venv
 VENV := .venv/bin
-SHELL = /usr/bin/env bash
+SHELL := /bin/bash
```
jamezpolley, line 6: The switch from = to := makes sense.

I'm not sure about the switch from "/usr/bin/env bash" to hard-coding "/bin/bash" I'm less sure about. I'm not sure about current OS X, but I know that older OS X versions used to ship with a really really old version of Bash, and it was common for users to install a newer version. In those cases, "/bin/bash" would pick the old version, "/usr/bin/env bash" would get the user's preferred version.

@benrfairless @simonzippy Could you check what "/usr/bin/env bash" gives for you? If you're both getting "/bin/bash" anyway, this change is probably okay.

Makefile
```diff
@@ -18,14 +22,11 @@ provisioning/.roles-installed: venv provisioning/requirements.yml
...
 
-ansible: venv roles ## Run Ansible on production
-	${VENV}/ansible-playbook --user=root --inventory-file=provisioning/hosts provisioning/playbook.yml
-
 help: ## This help dialog.
 	@IFS=$$'\n' ; \
 	help_lines=(`fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##/:/'`); \
-	printf "%-30s %s\n" "target" "help" ; \
-	printf "%-30s %s\n" "------" "----" ; \
+	printf "%-30s %s\n" "Target" "Description" ; \
```
jamezpolley, line 28: nice!

Makefile
```diff
@@ -1,13 +1,17 @@
...
+PYTHON_VERSION := $(shell cat .python-version 2>/dev/null || echo "python3")
 
-ALL: ansible
+all: help
 
 venv: .venv/bin/activate
 
 .venv/bin/activate: provisioning/requirements.txt
-	test -d .venv || virtualenv .venv
+	test -d .venv || virtualenv -p $(PYTHON_VERSION) .venv
```
jamezpolley, line 14: This comment applies equally to the old and new versions of this line, so it shouldn't stop this landing since the new version is definitely not worse than what it's replacing.

I would prefer to do this with "python${PYTHON_VERSION} -m venv .venv", rather than relying on virtualenv. the venv module is built-in and guaranteed to be present; virtualenv is a third-party tool which isn't guaranteed to be installed. Both will fail if python${PYTHON_VERSION} isn't installed and findable in $PATH; virtualenv may additionally fail even if python${PYTHON_VERSION} is installed and in the path if it happens to not be in a place that virtualenv is looking.

Corrollary: IMO `.python_-version` should say (for instance) `3.8` rather than what it currently says (`3.8.20`). That would make the command I've suggested come out as `python3.8` (which works) rather than `python3.8.20` (which intentionally doesn't work, as there shouldn't be any reason to prefer any particular sub-version of 3.8)

Again - my comment applies to both versions, I don't think this is a reason to reject this PR, the end result is better than the starting result even if I don't like either.

Makefile
```diff
@@ -37,25 +38,68 @@ help: ## This help dialog.
 		printf "%s\n" $$help_info; \
 	done
 
-up: ansible
-	${VENV} vagrant up local
+ # Ensure required Vagrant plugins are installed
+vagrant-plugins:
+	@plugins=$$(vagrant plugin list); \
```
jamezpolley: I misread this meaning that "@plugins" was going to be the list of plugins we want installed. Suggest renaming to something like "@installed_plugins" - and perhaps a new "@required_plugins" which contains the list to be installed, so that it's more clear what each does?

Makefile
```diff
@@ -37,25 +38,68 @@ help: ## This help dialog.
...
+			vagrant plugin install $$plugin; \
+		fi; \
+	done
+
+vagrant-up: venv roles vagrant-plugins ## launch local vagrant VM
+	vagrant up local
 
-provision: ansible
-	${VENV} vagrant provision local
+vagrant-provision: venv roles vagrant-plugins ## Provision local vagrant VM using ansible
```
jamezpolley, line 55: It doesn't make any difference, but seeing that "venv roles vagrant-plugins" appears a few times, I have a weak preference for abstracting this - maybe into a phony target, maybe into a variable, either way called something like "common" or "basic_setup".

It shouldn't make any appreciable difference when running make, but it gives a clearer signal to the reader that all of these things have the same dependencies, without the reader having to compare lines of text.

Procfile.production
```diff
@@ -1,2 +1,2 @@
-worker: rvm . do bundle exec sidekiq -C config/sidekiq.yml -e production
-faye: rvm . do bundle exec dotenv rackup sync.ru -E production
+worker: bash -c "until [ -e config/sidekiq.yml ]; do sleep 5; done; rvm . do bundle exec sidekiq -C config/sidekiq.yml -e production"
```
jamezpolley, line 1: I haven't seen this construction before, but I think I like it!

I had a longer comment here suggesting a change - but that was because I had missed the fact that this is no longer part of the `Makefile`. I've removed my comment. This is one of the reasons I prefer smaller changes - it's too easy for me to lose context on a bigger change like this.

README.md
```diff
@@ -42,23 +50,56 @@ which includes Docker Engine.
...
+
+* production-provision - Provision production using ansible
+* production-deploy - Deploy app to production
 
 Morph needs various services to run. We've made things easier for development by using docker
-to run Elasticsearch and the other services as well as the web container for the ruby on rails application itself.
+to run Elasticsearch and the other services.
 
-    docker compose up --build
+    make services-up
```
jamezpolley, line 91: Just adding a note that I like most of these comments. I hate the fact that I usually only comment when there's a problem - it gives the wrong impression :)

README.md
```diff
@@ -32,7 +32,15 @@ Read the [provisioning README](provisioning/README.md) for further details.
...
-Docker compose is used to provide a consistent development environment.
+### Support for Developers / CI
+
+Docker compose is used to provide redis, elasticsearch and mysql services as required for dev and CI
+(use SERVICES to specify which services to start if you don't want them all).
+
+vagrant is used to provide a local staging environment to test ansible provisioning and capistrano app deployment.
+
+If you don't want to set up ruby on your local host and/or have a different enough docker / mysql / redis version
+then use vagrant and /vagrant mapping of project root.
```
jamezpolley: I **think** that "/vagrant mapping of project root" means something like "Uncomment the `local.vm.synced_folder` in the `Vagrantfile`.

Could this be made more explicit?

Vagrantfile
```diff
@@ -17,7 +17,10 @@ Vagrant.configure("2") do |config|
     local.hostsupdater.aliases = ["faye.dev.morph.io", "api.dev.morph.io", "help.dev.morph.io"]
     local.vm.network :forwarded_port, guest: 22, host: 2200
     local.vm.network :forwarded_port, guest: 4443, host: 4443
-    local.vm.synced_folder ".", "/vagrant", disabled: true
+    # faye.dev.morph.io, dev.morph.io, api.dev.morph.io
+    local.vm.network :forwarded_port, guest: 443, host: 8443
+    # Use vagrant as both staging AND development!
+    # local.vm.synced_folder ".", "/vagrant", disabled: true
```
jamezpolley, line 23: Contrary to my earlier comment - I now see that this has "disabled: true" and is commented out. 

I'm not sure what I'd have to do to turn this on. Usually if something is commented out, uncommenting it turns it on. But on this case I'm uncommenting "disabled: true" - I'm un-disabling the disabled disabling. So that means that if I uncomment the line - nothing changes? Or maybe the default situation is to have this enabled, so having the line commented out means that it's enabled already, and I'd uncomment this if I don't want it? that means we've disabled the disabling. 

(sidebar - this isn't related to this change, but using names like "disabled" is a pet peeve of mine. [Don't use negatives](https://medium.com/@Cuadraman/why-to-stop-writting-negavite-code-af5ffb17195). )

I'd like a comment here explaining the default and how to get the other situation. If the default is disabled, then maybe the commented-out version should be the thing that enables this, so that simply uncommenting is all I need to do?

Vagrantfile
```diff
@@ -33,7 +36,21 @@ Vagrant.configure("2") do |config|
       tags = ENV["TAGS"].to_s.gsub(/[^A-Z0-9_]+/i, ",").split(",").reject { |s| s.to_s == "" }
       if tags.any?
         puts "INFO: Only running TAGS: #{tags.inspect}"
-        ansible.tags = tags if tags.any?
+        ansible.tags = tags
+      end
+      skip_tags = ENV["SKIP_TAGS"].to_s.gsub(/[^A-Z0-9_]+/i, ",").split(",").reject { |s| s.to_s == "" }
+      if skip_tags.any?
+        puts "INFO: Skipping TAGS: #{skip_tags.inspect}"
+        ansible.skip_tags = skip_tags
```
jamezpolley, line 44: If a tag is listed in both tags and skip_tags, what happens?

app/abilities/ability.rb
```diff
@@ -7,7 +7,5 @@ class Ability
   include CanCan::Ability
 
   sig { params(_user: T.nilable(Owner)).void }
-  # rubocop:disable Style/RedundantInitialize
-  def initialize(_user); end
-  # rubocop:enable Style/RedundantInitialize
+  def initialize(_user); end # rubocop:disable Style/RedundantInitialize
```
jamezpolley, line 10: 👍

app/helpers/scrapers_helper.rb
```diff
@@ -45,7 +45,7 @@ def auto_link_fallback(text)
   sig { params(text: String).returns(T::Boolean) }
   def url?(text)
     u = URI.parse(text)
-    u.scheme == "http" || u.scheme == "https"
+    %w[http https].include?(u.scheme)
```
jamezpolley, line 48: I don't like this. I *think* that it must do the same thing, but I had to think about it. The original version is easier to read.

Is there a reason why this is better and justifies the loss of readability?

app/lib/morph/docker_runner.rb
```diff
@@ -272,13 +273,18 @@ def self.docker_build_command(image, commands, dir, &block)
             block.call c
           end
         end
+        Rails.logger.debug { "Built docker container: #{image&.id || 'FAILED'}" }
```
jamezpolley, line 276: 👍

app/lib/morph/docker_runner.rb
```diff
@@ -306,11 +312,13 @@ def self.compile(image, repo_path, &block)
             "ENV NODE_TLS_REJECT_UNAUTHORIZED 0",
             # In development to debug what the buildpacks are doing it can be useful to turn on
             # the TRACE environment variable below by uncommenting the line
-            # "ENV TRACE true",
+            ENV["TRACE_BUILD"] ? "ENV TRACE true" : "",
             "RUN /bin/herokuish buildpack build"
           ],
           dir, &block
         )
+        Rails.logger.debug { "Compiled #{repo_path} image with app prerequisites: #{app_image&.id || 'FAILED'}" }
```
jamezpolley, line 320: 👍

app/lib/morph/docker_utils.rb
```diff
@@ -203,7 +203,8 @@ def self.docker_build_from_dir(dir, connection_options, &block)
       end
     # This exception gets thrown if there is an error during the build (for
     # example if the compile fails). In this case we just want to return nil
-    rescue Docker::Error::UnexpectedResponseError
+    rescue Docker::Error::UnexpectedResponseError => e
```
jamezpolley, line 206: 👍

jamezpolley (dismissed):
Mirroring [Matthew's comment](https://github.com/planningalerts-scrapers/multiple_masterview/pull/18#issuecomment-2664230911), I'm not a fan of big PRs like this. It does far too much for me to effectively review as a whole.

I've been going through it trying to focus *only* on the "occasional" bits that I understand and have an opinion on; but even so, it's taken me an hour to review 12/58 files. I have to stop now because I don't have enough hours in the day. I suspect it's going to take a while to work through my comments and figure out which ones need responses, too. 

As Matthew was (at least trying to, I think, if I've understood correctly) say, I'd much rather see this is a series of smaller self-contained PRs so that each change can be considered independently. 

(to be clear though - I'm not asking for this PR to be broken up, unless you think that would be easier. It's hear now, let's get it moving forward, but I'd be happier not to see goliath changes like this in future)

Dockerfile
ianheggie-oaf: I removed it because:
A. it is unnecessary because bundler will install the "BUNDLED WITH" version listed in Gemfile.lock if it differs from what is running (since v1.13.x aug 2016, and the default for this ruby is 2.1.4)
B. it would have been a bad idea anyway, as it wasn't version constrained, and would have broken the Dockerfile, eg:
```
ianh@ben:.../my_repos/morph (main *)$  gem install bundler
Fetching bundler-2.7.2.gem
ERROR:  Error installing bundler:
	The last version of bundler (>= 0) to support your Ruby & RubyGems was 2.4.22. Try installing it with `gem install bundler -v 2.4.22`
	bundler requires Ruby version >= 3.2.0. The current ruby version is 2.7.6.219.
```

Gemfile
ianheggie-oaf: gem "elasticsearch", "~> 7" would install any version after 7.0 onwards including breaking changes eg the latest 

[elasticsearch](https://rubygems.org/gems/elasticsearch) v9.1.2 with Required Ruby Version: >= 2.6

gem "elasticsearch", "~> 7.17" matches the following actual versions:

- [7.17.11](https://rubygems.org/gems/elasticsearch/versions/7.17.11) June 03, 2024 (15 KB)
- [7.17.10](https://rubygems.org/gems/elasticsearch/versions/7.17.10) November 07, 2023 (15 KB)
- [7.17.9](https://rubygems.org/gems/elasticsearch/versions/7.17.9) September 26, 2023 (15 KB)
- [7.17.8](https://rubygems.org/gems/elasticsearch/versions/7.17.8) September 05, 2023 (15 KB)
- [7.17.7](https://rubygems.org/gems/elasticsearch/versions/7.17.7) October 25, 2022 (15 KB)
- [7.17.1](https://rubygems.org/gems/elasticsearch/versions/7.17.1) February 28, 2022 (15 KB)
- [7.17.0](https://rubygems.org/gems/elasticsearch/versions/7.17.0) February 01, 2022 (15 KB)

Makefile
ianheggie-oaf: @james, @benrfairless, @simonzippy - If we need to do that for Mac support then OK lets do that, and do it consistently. Noting that the idea that the same script will run different bash versions with different behaviours on different peoples machines depending what homebrew installs they did sends shivers down my spine! Not that we should depend on any behaviour fancy enough it has changed on recent enough machines.

Personally I am used to the convention of hard-coding sh and bash like linux and unix does internally to protect against hacking - evaluating $PATH for system commands was an attack I heard some fellow Uni students where using bask in 1981 to gain root permission. So if we don't need to, lets not!

Makefile
ianheggie-oaf: very minor thing - everything else (ruby, gems etc) is locking down to an exact version rather than MAJOR.MINOR (mainly to be able to force a specific patch where needed) - if there was a venv command that accepted the full version then I don't care either way - if not I would prefer to keep the concept that everyone has the same for everything.

app/helpers/scrapers_helper.rb
ianheggie-oaf: github has been configured to reject PRs that don't pass the lint rules, and the [Style/MultipleComparison](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MultipleComparison) rule removes code duplication.

I am mildly in favour since anywhere there is code duplication, there could be bugs from subtle unnoticed differences that may not be found in simple testing (eg if it was miss-spelt ` u.scheme == "http" || u.schema == "https"` testing would not discover this bug unless `u.scheme` was not `"http"`

anyway rubyocop thinks its an automatically correctable offence, and presumably so does github lint, because its using the same rubocop rules eg:
```
Offenses:

app/helpers/scrapers_helper.rb:48:5: C: [Corrected] Style/MultipleComparison: Avoid comparing a variable with multiple items in a conditional, use Array#include? instead.
    u.scheme == "http" || u.scheme == "https"
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
app/helpers/scrapers_helper.rb:48:5: C: [Corrected] Style/WordArray: Use %w or %W for an array of words.
    ["http", "https"].include?(u.scheme)
    ^^^^^^^^^^^^^^^^^
```

README.md
ianheggie-oaf: I have clarified the README.md line in commit 132d80e4 - note Vagrantfile was already updated to reenable the default behaviour in commit f0e37d10c

Vagrantfile
ianheggie-oaf: This mess is required if you want to disable the default synched folder "/vagrant" - you have to have a config line that overwrites the default and then say it is disabled.  In this case I no longer want to disable the default mapping.

Vagrantfile
ianheggie-oaf: Apparently nothing much, eg:
```
$ TAGS=ruby SKIP_TAGS=ruby vagrant provision 
INFO: Only running TAGS: ["ruby"]
INFO: Skipping TAGS: ["ruby"]
==> local: [vagrant-hostsupdater] Checking for host entries
==> local: [vagrant-hostsupdater]   found entry for: 192.168.56.2 dev.morph.io
==> local: [vagrant-hostsupdater]   found entry for: 192.168.56.2 faye.dev.morph.io
==> local: [vagrant-hostsupdater]   found entry for: 192.168.56.2 api.dev.morph.io
==> local: [vagrant-hostsupdater]   found entry for: 192.168.56.2 help.dev.morph.io
==> local: Running provisioner: ansible...
    local: Running ansible-playbook...
PYTHONUNBUFFERED=1 ANSIBLE_FORCE_COLOR=true ANSIBLE_HOST_KEY_CHECKING=false ANSIBLE_SSH_ARGS='-o UserKnownHostsFile=/dev/null -o IdentitiesOnly=yes -o ControlMaster=auto -o ControlPersist=60s' ansible-playbook --connection=ssh --timeout=30 --limit="local" --inventory-file=/home/ianh/Projects/OpenAustralia/my_repos/morph/.vagrant/provisioners/ansible/inventory -v --tags=ruby --skip-tags=ruby provisioning/playbook.yml
Using /home/ianh/Projects/OpenAustralia/my_repos/morph/ansible.cfg as config file

PLAY [all] *********************************************************************

PLAY [all] *********************************************************************

TASK [Gathering Facts] *********************************************************
[DEPRECATION WARNING]: Distribution Ubuntu 16.04 on host local should use 
/usr/bin/python3, but is using /usr/bin/python for backward compatibility with 
prior Ansible releases. A future Ansible release will default to using the 
discovered platform python for this host. See https://docs.ansible.com/ansible/
2.9/reference_appendices/interpreter_discovery.html for more information. This 
feature will be removed in version 2.12. Deprecation warnings can be disabled 
by setting deprecation_warnings=False in ansible.cfg.
ok: [local]

PLAY [all] *********************************************************************

TASK [Gathering Facts] *********************************************************
ok: [local]

PLAY [development] *************************************************************

TASK [Gathering Facts] *********************************************************
ok: [local]

PLAY RECAP *********************************************************************
local                      : ok=3    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0 
```

Makefile
ianheggie-oaf: Thanks

Makefile
ianheggie-oaf: Fixed in 9b6e6bd1

Makefile
ianheggie-oaf: A borderline flip of the coin choice to my mind. venv and roles are different prerequisites, so I feel explicit is probably better for clarity especially since vagrant-plugins is also separate and not always required with the other two.

Procfile.production
ianheggie-oaf: Possibly because historically I have authored more code changes than reviewed, and I used different tools it didn't affect me as much ;)

README.md
ianheggie-oaf: I appreciate the feedback

ianheggie-oaf:
@jamezpolley - back to you to review please. I did find it hard to actually navigate to what githib was counting as "requested change" vs other comments, so I am starting to appreciate why medium and larger are not as convenient on this platform. (And I replied to your comments, and added changes as requested) - when this is accepted I will need to either merge the changes into the 3 PRs that are based on it or rebase them - which do you prefer?.

Makefile
Br3nda: i'm used to `#!/bin/bash`, but for things that move around, like `#!/usr/bin/env python`.
Let's go for the ubuntu standard?

Gemfile
Br3nda: cool, i understand this is saying 7.17 but less than 8.0

Makefile
```diff
@@ -37,25 +38,68 @@ help: ## This help dialog.
...
+	rm -rf .venv provisioning/.roles-installed tmp/cache
+
+clobber: clean ## Remove everything including logs
+	rm -f log/*.log
+
+docker-clean: services-down ## Remove all Docker resources INCLUDING databases in volumes
+	docker system prune -af --volumes
+
+share-web: ## Share web server on port 3000 to the internet (use PORT=N to use an alternative port)
+	ngrok http ${PORT:-3000}
```
Br3nda, line 104: :sparkles: looks useful

Procfile.production
Br3nda: What uses these Procfiles? I've seen them used in Heroku, but not sure what we're using it for here. Is it vestigial or still in use?

Br3nda: i think the answer is in our ansible that's also in this repo. So answering my own question (thanks Ian)

Br3nda (approved):
Approving -- but noting we agreed to make a new `staging` branch from this branch and sent PRs to that for a while until prod deploy time.

Vagrantfile
Br3nda: Ansible docs say:

Skipping always takes precedence over explicit tags, for example, if you specify both --tags and --skip-tags the latter has precedence. For example --tags tag1,tag3,tag4 --skip-tags tag3 will only run tasks tagged with tag1 or tag4, but not with tag3, even if the task has one of the other tags.

app/helpers/scrapers_helper.rb
Br3nda, line 48: maybe this works for both of you?
```suggestion
    u.scheme.match(/^https?$/)
```

Issue #1363, opened 2025-09-30
Add a html coverage report


ianheggie-oaf:
Currently morph only has a JSON Coverage report and it indicates 1956 / 3618 LOC (54.06%) covered.

Add html coverage report so you can see what is not covered by tests and add tests for critical areas that are not yet covered.

Issue #1353, opened 2025-09-30
Update morph to Rails 6.0.x


ianheggie-oaf:
Update morph to Rails 6.0.x, deploy to staging server for final QA and then production.

Note: We have changed the default branch to staging and am merging to there for the time being. staging is currently at my working_dev_example branch as of Mon 4am 10 Nov 2025.

Includes:
* https://github.com/openaustralia/morph/pull/1411
* https://github.com/openaustralia/morph/pull/1406

Branches merged in to upgrade_to_rails_6_0 branch:
* trim_log_lines_table - so there is space on production for sql backups to work, which depends on - _resolved lib/tasks/db.rake conflict_
  * includes deploy_BRANCH_or_current_branch - to be able to deploy branches before PRs accepted to staging and production
* add_db_backup_task - to do a manual backup 
* add_staging_target - to be able to deploy to staging - _Resolved numerous conflicts_
* clarify_ngix_logging - to assist in debugging issues as logs where being flooded with probes

TBD:

  change_from_rvm_to_mise
  get_rvm_working

ianheggie-oaf:
Agreed with @benrfairless to Move Fast And Break Things as its believed to be low risk and good return. I:
* backed up mysql database to deploy@morph:/var/www/current/db/backups/backup.sql.zst and downloaded to my dev system
* backed up redis:
```
cdmorph
$ du -ms db/scrapers/data/ db/backups/*
119684	db/scrapers/data/
4125	db/backups/backup.sql.zst
$ deploy@morph:/var/www/shared$ df -m .
Filesystem     1M-blocks   Used Available Use% Mounted on
/dev/root         643563 404916    205967  67% /
$ r console
>  redis = Redis.new
>      redis.save
> exit
```
On my machine :
```
scp root@morph:/var/lib/redis/dump.rdb ,backups/
scp deploy@morph:/var/www/current/db/backups/backup.sql.zst ,backups/
rsync -avz deploy@morph:/var/www/shared/db/scrapers/data/ ,backups/db/scrapers/data/ # left running - will take another hour
```
FYI - Confirmed previous deploys:
```
deploy@morph:/var/www/releases$ egrep . */REV*
20251107231249/REVISION:73d4091d1e8b4502680edb1a34afe86c50feabfa
20251107231911/REVISION:73d4091d1e8b4502680edb1a34afe86c50feabfa
20251107233714/REVISION:73d4091d1e8b4502680edb1a34afe86c50feabfa
[...]

PR #1367, opened 2025-10-11
Upgrade morph to rails 6.0 (initial work)

ianheggie-oaf:
Upgrades to rails 6.0

Builds on
* https://github.com/openaustralia/morph/pull/1366

Related to:
* https://github.com/openaustralia/morph/issues/1363
* https://github.com/openaustralia/morph/issues/1353


Gemfile
```diff
@@ -91,14 +94,14 @@ gem "stripe"
...
-gem "bootsnap", require: false
+gem "bootsnap", "~> 1.4", require: false
 
 # For type checking
 gem "sorbet-static-and-runtime"
 
 # Psych just seems to be giving us trouble
 # TODO: Figure out what's going on here
-gem "psych", "< 4"
+gem "psych" # , "< 4"
```
Br3nda: Is this still a TODO? Should we lock this down more?

Gemfile
```diff
@@ -187,7 +191,7 @@ gem "jbuilder"
 # We're only including sprockets here so we can lock it to an older version
 # For upgrading: https://github.com/rails/sprockets/blob/070fc01947c111d35bb4c836e9bb71962a8e0595/UPGRADING.md#manifestjs
 # TODO: Upgrade to sprockets version 4 and remove the line below
-gem "sprockets", "~> 3"
+gem "sprockets", "~> 4.0"
```
Br3nda, line 192: is the comment above still valid?

app/models/scraper.rb
```diff
@@ -35,7 +35,7 @@ class Scraper < ApplicationRecord
   has_many :api_queries, dependent: :delete_all
 
   validates :name, presence: true, format: { with: /\A[a-zA-Z0-9_-]+\z/ }
-  validates :name, uniqueness: { scope: :owner }
+  validates :name, uniqueness: { scope: :owner, case_sensitive: false }
```
Br3nda, line 38: are there existing conflicts in our DB same name different case?

config/database.yml.example
```diff
@@ -12,13 +12,21 @@ development: &development
...
 # Do not set this db to the same as development or production.
 test:
-  <<: *development
+  adapter: mysql2
+  encoding: utf8mb3 # TODO: update to utf8mb4
+  collation: utf8mb3_unicode_ci # TODO Update to utf8mb4_general_ci (5.7 compatible) or default in 8.0
+  username: <%= ENV.fetch("MYSQL_USER", "scraping").inspect %>
+  password: <%= ENV.fetch("MYSQL_ROOT_PASSWORD", "").inspect %>
+  host: <%= ENV.fetch("MYSQL_HOST", "localhost").inspect %>
+  pool: 5
```
Br3nda: are you running 5 parallel tests? Not sure why we use a pool here

config/initializers/cookies_serializer.rb
```diff
@@ -2,4 +2,5 @@
 
 # Specify a serializer for the signed and encrypted cookie jars.
 # Valid options are :json, :marshal, and :hybrid.
-Rails.application.config.action_dispatch.cookies_serializer = :marshal
+# FIXME: Change to :json 1 hour after after next deploy (for seamless upgrade of user session cookies)
+Rails.application.config.action_dispatch.cookies_serializer = :hybrid
```
Br3nda, line 6: sounds like this needs to be in the deploy procedures.

config/initializers/inflections.rb
```diff
@@ -15,6 +15,3 @@
 #   inflect.acronym 'RESTful'
 # end
 
-ActiveSupport::Inflector.inflections(:en) do |inflect|
```
Br3nda, line 18: What was this all about?

Gemfile
ianheggie-oaf: Fixed: Actually I was too early in releasing this lock (the fix was in rails 6.1 not 6.0) reverted and explained in commit a78717cf

Gemfile
ianheggie-oaf: Updated comment in commit dde0e740 to clarify

app/models/scraper.rb
ianheggie-oaf: No, there are no conflicts - I ran the following to check:
```
deploy@morph:/var/www/current$ echo 'select owner_id,name from scrapers' | r db -p | tr '[A-Z]' '[a-z]' | sort | uniq -d
mysql: [Warning] Using a password on the command line interface can be insecure.
deploy@morph:/var/www/current$ 
```

Note: 
* "github repository names are, in fact, not case-sensitive" https://github.com/github/docs/issues/32838
* Rails 6.0 changes the case sensitivity default for uniqueness check

config/database.yml.example
ianheggie-oaf: These lines are no longer added: Reverted expanding out a yaml alias in commit af29bf37 - problem resolved by instead reverting to psych 3.x gem in earlier commit

config/initializers/inflections.rb
ianheggie-oaf: I accidentally let the app:update task remove this when I should not have - there is a Plan called "Hero". I have reverted the change and added a spec to make sure it is kept in the future in commit 8c62df8d

ianheggie-oaf:
> the things that usually trips us up on the other rails apps is getting that ruby version changed. (they all share a ruby version with a symlink)

Morph explicitly sets the ruby version for that site in the nginx site config:
```
root@morph:/etc/nginx/sites-enabled# egrep ruby *
morph.io:  passenger_ruby /home/deploy/.rvm/gems/ruby-2.7.6/wrappers/ruby;
```

This comes from the ansible rules:
```
ianh@ben:.../my_repos/morph (upgrade_to_rails_6_0)$->(2) ag passenger_ruby
provisioning/roles/morph-app/templates/nginx.conf
65:	passenger_ruby /usr/bin/ruby;

provisioning/roles/morph-app/templates/sites/morph.io
6:  passenger_ruby /home/deploy/.rvm/gems/ruby-{{ ruby_version }}/wrappers/ruby;
```

config/initializers/cookies_serializer.rb
ianheggie-oaf: I have added a commit to a later rails branch to update this to :json so i don't forget

ianheggie-oaf (commented):
Thanks for the review, that helped. You highlighted some areas that needed to be changed - please check it makes sense now. I will know to add terse comments why changes where needed in future to make the review go faster.

Br3nda (approved):

Issue #1354, opened 2025-09-30
Update morph from Rails 6.0 to Rails 6.1.x


ianheggie-oaf:
Update morph from Rails 6.0.6.1 to Rails 6.1.x, deployed to the current production server. This is Phase 1 of [`doc/UPGRADE_PATH.md`](https://github.com/openaustralia/morph/blob/main/doc/UPGRADE_PATH.md) (revised 20 August 2026), which reinstated the 6.1 stepping stone after this issue was earlier closed in favour of a direct 6.0 to 7.0 jump.

Part of epic #1351. Blocks #1355 (Rails 7.0 + Ruby 3.0.7).

## Already landed (Phase 1 precursors)

- #1438 unsafe sign-out redirect fix (PR #1517)
- Sentry gems alongside Honeybadger (PR #1518, advances #1474). With no staging environment, Sentry is the soak instrument for this and every later hop.

## Remaining work

1. **Mine `chore/rails_61` (PR #1423), do not rebase it.** Stripped of regenerable noise (RBIs, schema.rb, lockfiles) it is roughly 15 substantive changes. Mining list: the Gemfile bump, `add_template_helper` to `helper` in `AlertMailer`, the `errors.keys` replacement in `Domain`, and the spec adjustments. Skip its spring workaround (`main` is on spring 4, which should carry the fix; verify) and its Active Storage migrations (#1497 deliberately removed that config). Delete the branch once mined.
2. Run `rails app:update` fresh, then adopt the 6.1 defaults in a single PR with one commit per flag in `new_framework_defaults_6_1.rb`, ending with `config.load_defaults 6.1`. The only flag needing real thought is `cookies_same_site_protection = :lax`: the GitHub OAuth callback is a top-level redirect so Lax should be fine, but verify sign-in explicitly after deploy.
3. Unpin `psych` (Rails 6.1 carries the Psych 4 fix).
4. Regenerate Sorbet RBIs (`bundle exec tapioca gem`); fix deprecations until the #1498 deprecation-raise guard passes.
5. Deploy to production in a low-usage window. There is no staging environment, so `cap production deploy:rollback` is the fallback and Sentry is the monitor.

## Exit criteria

- CI green including Brakeman and bundle-audit
- Zero Rails deprecation warnings (enforced by the #1498 guard)
- `srb tc` clean after RBI regeneration
- GitHub sign-in verified after deploy (SameSite=Lax)
- About a week of production soak on Sentry before starting #1355

[...]

PR #1368, opened 2025-10-11
WIP2 - Upgrade to rails 6.1 [from fork: replaced]

ianheggie-oaf:
Upgrades morph to rails 6.1

Builds on
* https://github.com/openaustralia/morph/pull/1367

Related to:
* https://github.com/openaustralia/morph/issues/1354


app/models/webhook.rb
```diff
@@ -6,7 +6,7 @@ class Webhook < ApplicationRecord
 
   belongs_to :scraper
   has_many :deliveries, class_name: "WebhookDelivery", dependent: :delete_all
-  validates :url, presence: true, url: true, uniqueness: { scope: :scraper }
+  validates :url, presence: true, url: true, uniqueness: { scope: :scraper, case_sensitive: true }
```
Br3nda, line 9: I think case_sensitive is the default anyway.

bin/rails
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
+# load File.expand_path("spring", __dir__) # disable spring as it was causing Rails::Application is abstract, you cannot instantiate it directly in railties gem
 APP_PATH = File.expand_path('../config/application', __dir__)
-require_relative '../config/boot'
-require 'rails/commands'
+require_relative "../config/boot"
+require "rails/commands"
```
Br3nda, line 5: so rubocop made us do this?

bin/setup
```diff
@@ -9,24 +8,25 @@ def system!(*args)
...
   # Add necessary setup steps to this file.
 
   puts '== Installing dependencies =='
   system! 'gem install bundler --conservative'
   system('bundle check') || system!('bundle install')
 
-  # Install JavaScript dependencies if using Yarn
-  # system('bin/yarn')
+  # Install JavaScript dependencies
+  system! 'bin/yarn'
```
Br3nda, line 21: Are we sure we need this? Did we add yarn?

bin/spring
```diff
@@ -1,15 +1,14 @@
...
-  if match = Bundler.default_lockfile.read.match(/^GEM$.*?^    (?:  )*spring \((.*?)\)$.*?^$/m)
-    Gem.paths = { "GEM_PATH" => [Bundler.bundle_path.to_s, *Gem.path].uniq }
-    gem "spring", match[1]
+  # Load Spring without loading other gems in the Gemfile, for speed.
+  Bundler.locked_gems&.specs&.find { |spec| spec.name == "spring" }&.tap do |spring|
+    Gem.use_paths Gem.dir, Bundler.bundle_path.to_s, *Gem.path
+    gem "spring", spring.version
     require "spring/binstub"
+  rescue Gem::LoadError
+    # Ignore when Spring is not installed.
```
Br3nda, line 12: did this change come with rails? Is so, all good.

could hide errors from people debugging one day

config/boot.rb
```diff
@@ -1,4 +1,8 @@
 ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
 
 require 'bundler/setup' # Set up gems listed in the Gemfile.
+
+# TODO: Remove the following rails 6.0,6.1 fix when upgrading to Rails 7.0+ (LoggerThreadSafeLevel removed)
```
Br3nda, line 5: So, it's removed in your next PR? all good

config/cable.yml
```diff
@@ -2,7 +2,7 @@ development:
   adapter: async
 
 test:
-  adapter: async
+  adapter: test
```
Br3nda, line 5: Why did we change this?

db/migrate/20131223033729_create_scrapers.rb
```diff
@@ -1,7 +1,7 @@
 class CreateScrapers < ActiveRecord::Migration[4.2]
   def change
     create_table :scrapers do |t|
-      t.string :name, null: false
+      t.string :name, default: "", null: false
```
Br3nda, line 4: This is an old migration that already run, so why are we modifying it?

provisioning/group_vars/all.yml
```diff
@@ -25,6 +25,12 @@ apt_packages:
   - sqlite3
   - zlib1g-dev
 
+bash_alias_lines:
+  - "alias cdmorph='if [ \"$USER\" = \"deploy\" ]; then cd /var/www/current; export RAILS_ENV=production; env | egrep RAILS_ENV; else cd /vagrant ; echo \"development. sudo -u deploy -i for production!\"; fi'"
+  - 'alias be="bundle exec"'
+  - 'alias r="bundle exec rails"'
+  - 'alias rk="bundle exec rake"'
+
```
Br3nda, line 34: Is that where we install on prod?  (okay, I  checked it is) `/var/www/...`
We're using `/srv/` on other deploys.

Br3nda (commented):
added comments. none are blocking comments. Happy to approve, just checking for any replies from @ianheggie-oaf  before that

provisioning/group_vars/all.yml
Br3nda: To be clearer, i'm not proposing changing this on our existing server. But how about we change to `/srv/` when we build the new server, to make them all the same.

Br3nda: these do look like personal aliases, but I can't see the harm in having them.

app/models/webhook.rb
ianheggie-oaf: A deprecation warning was introduced in Rails 6.0, indicating a change in behavior for Rails 6.1 and later versions. In Rails 6.1 and subsequent versions, the default behavior for validates_uniqueness_of changed to case-insensitive comparison.

THEREFOR I started making the choice explicit (and confirming which it should be)

bin/rails
ianheggie-oaf: I forget if `rails app:update` did it or `rubocop -a` - probably rubocop

bin/setup
ianheggie-oaf: I left it as app:update wanted, as otherwise i would be fighting with app:update on each update

bin/spring
ianheggie-oaf: came from rails from memory - I don't directly call bin/spring so I wouldn't have had a reason to change

config/cable.yml
ianheggie-oaf: changed by app:update

db/migrate/20131223033729_create_scrapers.rb
ianheggie-oaf: In commit aff821df "Fix create scraper migration so it matches actual production" - I changed this so doing a db:migrate from a fresh DB results in the same schema as production actually has!

provisioning/group_vars/all.yml
ianheggie-oaf: > To be clearer, i'm not proposing changing this on our existing server. But how about we change to `/srv/` when we build the new server, to make them all the same.

Added as issues as there are a number of places that are affected:
* https://github.com/openaustralia/morph/issues/1382
* https://github.com/openaustralia/infrastructure/issues/267

PR #1393, opened 2025-11-11
Fix backup database script

ianheggie-oaf:
This is the fix I applied manually to production to get backup database script working.

I also cleared out the backup files that where years old because the config has MAX_AGE=6M and it was being ignored.


jattree (approved):
Only small suggestion: the script currently calls mysqldump and mysql as --user=root without referencing a password source. If the host ever requires credentials, pulling them from /root/.my.cnf (or env vars) would make this more robust.

Other than that, this looks good to me.

PR #1394, opened 2025-11-13
Upgrade to rails 6.0 [replaced by branch PR]

ianheggie-oaf:
Includes work done on branches and resolving merge conflicts into upgrade_to_rails_6_0 branch - includes these PRs:
* https://github.com/openaustralia/morph/pull/1393
* https://github.com/openaustralia/morph/pull/1389
* https://github.com/openaustralia/morph/pull/1396
* https://github.com/openaustralia/morph/pull/1387

Provides:
  * https://github.com/openaustralia/morph/issues/1353
  
Demo site (requires domain to be added to github app): https://morph-staging.thesite.info/

I did the rails 6.0 upgrade on my dev system, then resolved deployment and QA issues in separate, as independent as possible branches. Unfortunately I ended up with some merge conflicts because of prior accepted PRs. Given limited remaining hours, rather than make a long line of PRs and probably have to rebase after each one, I merged all relevant branches into this (upgrade_to_rails_6_0) as it is the reason for all the separate work (along with merging in the production hotfixes).

* use_site_domain_in_views - uses sites domain (eg morph.io or whatever you set when deploying)
  * Just like ben's (?) request to have hostname shown when ssh to staging, this provides feedback when you are on a staging server
* fix_backup-database_script - fixes the copying of local sql backups to s3
  * https://github.com/openaustralia/morph/issues/1392
* production which had hotfixes:
  * replace_broken_daily_backup - fixzes the generation of local sql backups
    * https://github.com/openaustralia/morph/issues/1392
  * add_db_backup - to backup production before recent data clean up
  * trim_log_lines_table - clean up old run logs
    * https://github.com/openaustralia/morph/issues/1373
  * And based off working_dev_example (merged PR)
* main branch - including PRs merged since production was last deployed
  * Not sure why git didn't think it was merged, but I merged it for consistency
* clarify_nginx_login - separate access and error logs for each nginx virtual host to see the wood for the trees when diagnosing staging  
* add_staging_target - so I could deploy to staging without overwriting production settings
  * https://github.com/openaustralia/morph/issues/1388
* deploy_BRANCH_or_current_branch
  * https://github.com/openaustralia/morph/issues/1372
  * https://github.com/openaustralia/morph/issues/1386
* support_ed25519_ssh_keypairs - so i can deploy with my ed25519 keys
  * https://github.com/openaustralia/morph/issues/1372
* add_html_coverage_report - so i could get a feel for test coverage to check if it was safe to upgrade without adding more tests
[...]

PR #1395, opened 2025-11-14
Annotate models with a comment summarizing their current schema when db:migrate runs [replaced]

ianheggie-oaf:
Annotates models files with a comment summarizing their current schema when db:migrate runs using the https://github.com/ctran/annotate_models (annotate) gem.

I ended up adding annotate gem as I find it very useful on other projects and I really miss the models being annotated..

Based on:
* https://github.com/openaustralia/morph/pull/1394


Br3nda:
some git conflicts to resolve here

Issue #1405, opened 2025-12-04
Provision yarn for rails 6.0


ianheggie-oaf:
Deploy breaks on the assets:precompile task if yarn is not present even though it technically is not required by our configuration.

ianheggie-oaf:
Closed as PR is merged

PR #1406, opened 2025-12-04
Provision yarn for rails 6.0

ianheggie-oaf:
Resolves:
* https://github.com/openaustralia/morph/issues/1405

I have tested this locally using vagrant and it works for me


provisioning/requirements.yml
```diff
@@ -15,3 +15,4 @@
 - src: newrelic.newrelic-infra
   version: 0.6.0
 
+- src: https://github.com/angristan/ansible-yarn
```
Br3nda, line 18: this repo is quiet and old. Is it still a good source for this?

benrfairless: This role just runs the tasks listed here:
https://github.com/angristan/ansible-yarn/blob/master/tasks/main.yml

If it works, and those tasks remain valid, I don't think the age of the repo is a problem.

ianheggie-oaf: @brenda and @benrfairless - are you happy to approve? I dont think the age is an issue either.

benrfairless (approved):

Br3nda (approved):

Issue #1412, opened 2025-12-04
Add cap staging target


ianheggie-oaf:
**Problem**

Staging needs to be different from production. It is cleanest to set RAILS_ENV=staging so stuff that shouldn't be set in staging is automatically configured accordingly

**Desired solution**

Add a staging capistrano deploy target and rails environment and update the configuration files accordingly

**Alternatives I've considered**

Considered doing this via .env, but then you have to remember to do it.

PR #1413, opened 2025-12-04
Add cap staging target for rails 6.0

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/morph/issues/1412

## What does this do?

Add a staging capistrano deploy target and rails environment and update the configuration files accordingly

Also adds mailcatcher gem for development catching of emails.

## Why was this needed?

Staging needs to be different from production. It is cleanest to set RAILS_ENV=staging so stuff that shouldn't be set in staging is automatically configured accordingly


config/deploy.rb
```diff
@@ -1,24 +1,42 @@
 set :application, 'morph'
-set :repo_url, 'https://github.com/openaustralia/morph.git'
+set :repo_url, `git config --get remote.origin.url`.strip
```
Br3nda: ah, this means it uses whatever you have configured locally. Not sure how this is better.

Br3nda (changes requested):
can we keep it pointing at the OAF git repo?

config/deploy.rb
ianheggie-oaf: Reverted back and added a check since we have moved to branch based deployment:

```
set :repo_url, "https://github.com/openaustralia/morph.git"

expected_repos = %w[git@github.com:openaustralia/morph.git https://github.com/openaustralia/morph.git]
current_repo = `git config --get remote.origin.url`.strip
raise "Deploy from the openaustralia repo (#{expected_repos.join(', ')}) instead of a fork (#{current_repo})" unless expected_repos.include?(current_repo)
```

Br3nda (approved):

Issue #1414, opened 2025-12-04
Show domain in website so there is no confusion between production and staging for rails 6.0


ianheggie-oaf:
**Description**
To avoid mistakes, similar to Bens request for ssh prompt always show and use the current domain for display and links

**Desired solution**
Pass SERVER_NAME from .env into the app, and use it for email, urls and navbar domain

ianheggie-oaf:
Closed - PR was merged

PR #1415, opened 2025-12-04
Use actual server name rather than morph.io for staging for rails 6.0

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/morph/issues/1414

## What does this do?

Shows the current domain in views and uses it in urls and emails so there is no confusion between staging and production

## Why was this needed?

Too easy to make mistakes (similar reason to Ben's ssh prompt request)

## Notes

Used by PR:

* https://github.com/openaustralia/morph/pull/1421


app/helpers/site_helper.rb
```diff
@@ -0,0 +1,30 @@
+# typed: strict
+# frozen_string_literal: true
+
+# Helpers for site details
+module SiteHelper
+  extend T::Sig
```
Br3nda, line 6: Can we have some tests for the methods on here? So we can see that they do what we expect. (and we'll notice in the future if we break them)

Br3nda (changes requested):

app/helpers/site_helper.rb
ianheggie-oaf: Yes - app/helpers/site_helper.rb now has 100.0% lines covered. Overall COVERAGE:  54.17% -- 2033/3753 lines in 91 files

app/models/scraper.rb
```diff
@@ -40,6 +40,9 @@
 class Scraper < ApplicationRecord
   extend T::Sig
 
+  # Make skipping GitHub validations an explicit choice
+  class_attribute :skip_github_validations, default: -> { Rails.env.test? }
+
```
Br3nda, line 45: What does this do?

app/models/scraper.rb
```diff
@@ -375,15 +378,15 @@ def trim_log_lines
...
+    return if self.class.skip_github_validations ||
+              !Octokit.client.repository?(full_name)
 
     errors.add(:name, "is already taken on GitHub")
   end
 
   sig { void }
   def app_installed_on_owner
-    return if Rails.env.test?
+    return if self.class.skip_github_validations
```
Br3nda, line 389: Not sure what this change is about

app/models/scraper.rb
ianheggie-oaf: This establishes a class attribute that is set to the previously hard-coded test later in the code. It sets the default to what the code used to test, and the code down further tests this class attribute. THUS we can set or clear this when setting up test runs.

Test factory data won't pass github validation because we arn't using real values. Thus if we want to load that test data into development (for development with the data that test uses), we have to be able to disable the validations.

app/models/scraper.rb
ianheggie-oaf: Moves the hard coded test into a class attribute

Issue #1448, opened 2026-01-04
Update example Node.js scraper.js to heroku-24 and current language / library versions


ianheggie-oaf:

ianheggie-oaf:
Updated and passes spec see PR:
* https://github.com/openaustralia/morph/pull/1452

Issue #1447, opened 2026-01-04
Update example Perl scraper.pl to heroku-24 and current language / library versions


ianheggie-oaf:
Update example Perl scraper.pl to heroku-24 and current language / library versions

My heroku-18 example scraper works: https://morph.io/ianheggie-oaf/example_perl_scraper

For the time being it has been removed from Morph::Language::LANGUAGES_SUPPORTED, and spec/lib/morph/docker_runner_spec.rb checks and marks associated tests as pending if the language is not on the supported list.

Development branch: feature/update_perl_example_to_heroku_24

## heroku-24 FAILS

```
Injecting configuration and compiling...
[1G       [1G-----> Perl/PSGI app detected
[1G-----> Bootstrapping cpanm
```
## heroku-18 works
```
Injecting configuration and compiling...
Injecting scraper and running...
Running perl 5.026001 on Linux b772785537eb 6.14.3-x86_64-linode168 #1 SMP PREEMPT_DYNAMIC Mon Apr 21 19:47:55 EDT 2025 x86_64 x86_64 x86_64 GNU/Linux
5: Example Domain
4: Example Domain
3: Example Domain
```

Issue #1446, opened 2026-01-04
Update example Python scraper.py to heroku-24 and current language / library versions


ianheggie-oaf:
Update example Python scraper.py to heroku-24 and current language / library versions.

Python scrapers are currently broken: 
* https://github.com/openaustralia/morph/issues/1344

My example heroku-24 scraper is broken: https://morph.io/ianheggie-oaf/example_python_scraper
and it is attempting to run the wrong python version anyway (2.7 rather than 3.14)

Dev branch: feature/update_python_example_to_heroku_24

# heroku-24 fails
```
Injecting configuration and compiling...
       -----> Python app detected
-----> Using Python 3.12.8 specified in runtime.txt

 !     Error: Python 3.12.8 isn't available for this stack (heroku-24).
 !     
 !     For a list of the supported Python versions, see:
 !     https://devcenter.heroku.com/articles/python-support#supported-runtimes
```

Issue #1445, opened 2026-01-04
Update example PHP scraper.php to heroku-24 and current language / library versions


ianheggie-oaf:
Update example PHP scraper.php to heroku-24 and current language / library versions.

Consider improving code so it doesn't have deprecation warnings

My heroku-18 example scraper works on heroku-18: https://morph.io/ianheggie-oaf/example_php_scraper but not heroku-24

Development branch: feature/update_php_example_to_heroku_24

Should the example include use of composer?
* https://github.com/openaustralia/morph/issues/1204

For the time being it has been removed from Morph::Language::LANGUAGES_SUPPORTED, and spec/lib/morph/docker_runner_spec.rb checks and marks associated tests as pending if the language is not on the supported list.

Development branch: feature/update_php_example_to_heroku_24

## heroku-24 FAILS
```
Injecting configuration and compiling...
       -----> PHP app detected
----->  Bootstrapping...
 
 !     ERROR:  Failed to download minimal PHP for bootstrapping!
 !     
 !     This is most likely a temporary internal error. If the problem
 !     persists, make sure that you are not running a custom or forked
 !     version of the Heroku PHP buildpack which may need updating.
```

## heroku-18 OK
```
Injecting configuration and compiling...
Injecting scraper and running...
Running PHP 7.4.33 on Linux dac345017d9a 6.14.3-x86_64-linode168 #1 SMP PREEMPT_DYNAMIC Mon Apr 21 19:47:55 EDT 2025 x86_64 x86_64 x86_64 GNU/Linux
PHP Deprecated:  Array and string offset access syntax with curly braces is deprecated in /app/vendor/electrolinux/phpquery/phpQuery/phpQuery/phpQueryObject.php on line 1040
PHP Deprecated:  Array and string offset access syntax with curly braces is deprecated in /app/vendor/electrolinux/phpquery/phpQuery/phpQuery/phpQueryObject.php on line 1055
1: some value
11: Example Domain
```

ianheggie-oaf:
Removed from supported list - example partially updated, see PR:
* https://github.com/openaustralia/morph/pull/1452

PR #1452, opened 2026-01-06
Update morph to include heroku 24 platform

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/morph/issues/1440

## What does this do?

* Upgrades the support for a new platform (24)
* Adds a python fallback for limit_output (since the latest buildstep doesn't have system ruby)
* Changes "early_release" to be "heroku-24" to be consistent with its common meaning (latest platform available, with a connotation that it may still being worked on)
* Removes duplicated schema comments on models (noticed in passing)
* Added summary one-lines for models without them for consistency
* Updates scraper examples
  * https://github.com/openaustralia/morph/issues/1443 - examples
    * https://github.com/openaustralia/morph/issues/1444 - ruby
    * https://github.com/openaustralia/morph/issues/1448 - nodejs
    * https://github.com/openaustralia/morph/issues/1446 - python2 has been dropped and the wikiscraper gem needs to be updated for python3
      * I implemented a working python example similar to the ruby example so i could actually test it works (scrapes H1 headings from example.com) 
* I also  added issues for languages I started working on, but paused as it was taking too long. I dropped the following from the supported list (which removes them from documentation page)
  * https://github.com/openaustralia/morph/issues/1447 - perl dropped
  * https://github.com/openaustralia/morph/issues/1445 - php dropped    
* Updated the spec example scrapers to be explicit which platform they where for, moved many to heroku-24, and spread the ruby exampels across a few supported ruby versions
* Added comment on what versions where too new for the heroku-24 docker image james built
* Updated some string arrays to use `%w[]` syntax as recommended by rubocop/rubymine

## Why was this needed?

* Primarily so we could upgrade scrapers to avoid nokogiri security issues
* Also gets python3 and php working

Standard capistrano deploy


default_files/python/template/scraper.py
```diff
@@ -22,3 +22,49 @@
 # All that matters is that your final data is written to an SQLite database
 # called "data.sqlite" in the current working directory which has at least a table
 # called "data".
+
```
Br3nda, line 25: how about we ditch python 2 support entirely?

spec/spec_helper.rb
```diff
@@ -50,7 +50,7 @@
                           # * GitHub COVERAGE:  80.09% -- 2957/3692 lines in 97 files
                           # * Local COVERAGE:  80.44% -- 2970/3692 lines in 97 files
                           # SO Set it based on the CI coverage on GitHub!
-                          80.09
+                          80.05
```
Br3nda, line 53: :eyes:

Br3nda (approved):

Issue #1439, opened 2026-01-02
Make sure all of Ian's multiple_* repos and morph scrapers are all private


ianheggie-oaf:
Make sure all of Ian's multiple_* repos and morph scrapers are all private

Updated
* https://github.com/ianheggie-oaf/multiple_nsw_planning_lgas

Issue #1342, opened 2025-07-02
Allow developers to test private repos rather than insist they are public


ianheggie-oaf:
Currently to test using morph any of the private `multiple_*` scrapers I need to do the following, which defeats the purpose of them being private:

1. Clone the private repo to my account (which forces it to be private)
2. Create one or more public repos in my account (I append a suffix to give it a unique name, eg https://github.com/ianheggie-oaf/multiple_masterview-prs
3. On my system pull from my private repo, create branches for each authority I fix (so they can have separate PRs as required by OAF unwritten policy)
4. Push either a branch related to a PR or to a branch with multiple PR branches merged (my "prs" branch is the merge of all the PRs) to the **master** branch of the public repo
5. Have morph import the public repo - it only runs the default branch

BONUS if you allow a branch to be specified rather than always running the default branch!

PR #1458, opened 2026-01-13
Feature: allow webhook to include run stats

ianheggie-oaf:
## Relevant issue(s)

These are the issues that triggered the need for this change.

* https://github.com/openaustralia/morph/issues/1439
* https://github.com/openaustralia/morph/issues/1342

## What does this do?

As documented in app/views/documentation/webhooks.html.haml: The following UPPERCASE strings will be replaced with the current run details in the url:

  * ADDED: Records added count
  * AUTO: "true" if an automatic daily run, otherwise "false"
  * COMMIT: Full git revision
  * OUTCOME: "success" if run finished successfully, otherwise "failed"
  * REMOVED: Records removed count
  * REVISION: Abbreviated git revision (e.g. "c9fab2s7")
  * RUN_TIME: Seconds taken to run (eg "603" is 10 minutes and 3 seconds)
  * SCRAPER: Full scraper name, eg "owner-name/repo-name"
  * STATUS_CODE: Process exit code (0 means success)

Specs have been added to check the new functionality, as well as test if the webhook fails (previously uncovered check).

## Why was this needed?

My [plannies mate](https://plannies-mate.thesite.info/) site was unable to access run details for private repos since I made all of my multiple*-prs repos private. (which I did after I was given access to the "private" flag for morph).

**The time to do this was a donation to OAF**

Standard capistrano deploy.

## Notes to reviewer (Optional)

This shouldn't be a security issue in my opinion, since:

* For public scrapers, you don't need to login to see this data
* For private scrapers (our multiple_* scrapers), the same person who has access to [settings](https://morph.io/ianheggie-oaf/multiple_atdis-prs/settings) to set the webhook url, also has access to the [scraper page](https://morph.io/ianheggie-oaf/multiple_atdis-prs), which gives a red/green outcome line and lists the run history.
* No personally identifying information is included
* The values are safe for use, being constrained to either numbers, specific values, or in the case of scraper name, constrained to be url friendly by github and used by us as part of our url
  * Note: "/" has been valid in the query for the last 20 years [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4)

### DIFF
--- a/app/views/documentation/webhooks.html.haml
+++ b/app/views/documentation/webhooks.html.haml
@@ -33,6 +33,7 @@
   * REMOVED: Records removed count
   * REVISION: Abbreviated git revision (e.g. "c9fab2s7")
   * RUN_TIME: Seconds taken to run (eg "603" is 10 minutes and 3 seconds)
+  * SCRAPER: Full scraper name, eg "owner-name/repo-name"
   * STATUS_CODE: Process exit code (0 means success)
 
   If there’s something specific you’d like included in the webhook payload
--- a/app/views/documentation/webhooks.html.haml
+++ b/app/views/documentation/webhooks.html.haml
@@ -24,6 +24,17 @@
   You can send information through the request
   by putting it in the webhook URL,
   either as the path or as GET parameters.
+  The following UPPERCASE strings will be replaced with the current run details in the url:
+
+  * ADDED: Records added count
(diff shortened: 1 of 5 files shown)


benrfairless (approved):
LGTM (but I admit that I'm not an expert - I'm very much trusting you with this).

Issue #1403, opened 2025-12-04
Update Cron to reduce DB size weekly/monthly


benrfairless:
Run the task created in #1373 weekly/monthly to keep things up to date.

PR #1471, opened 2026-03-30
Expand db:stats rake task to include table + index size

### LINKED ISSUE #1470: Add size to db:stats report
**Feature**

Add size to db:stats task so we can see what effect the cleanup has

**Describe the solution you'd like**

Copy expanded db:stats developed on another repo

**Additional context**

Its a standard util I add whenever I care about DB size

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/morph/issues/1470 which is child of
* https://github.com/openaustralia/morph/issues/1403

## What does this do?

Adds size to db:stats rake task

## Why was this needed?

So we know effect of cleanup

Deploy using capistrano as normal

## Notes to reviewer (Optional)

Setting EXACT_COUNT gives exact table count but takes more time - normally not required

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -16,11 +16,12 @@ namespace :db do
     db_name = ActiveRecord::Base.connection.current_database
     conn    = ActiveRecord::Base.connection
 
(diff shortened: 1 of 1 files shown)


benrfairless (approved):
Approving on the assumption you have tested this and we can roll it back if needed.

Issue #1351, opened 2025-09-29
Update Morph Software versions to obtain security updates [Epic]


ianheggie-oaf:
Strategy I am executing to keep systems in support: 

Upgrade Rails when there is 2-6 months left of support to second latest minor version.
When Rails is upgraded, upgrade as needed so everything else has at support at least as long as Rails to reduce number of upgrades and complies with Rails / gem min version requirements. (Check software from non Ubuntu repositories as well)
When upgrading, choose versions recommended for use with that Rails version unless a later version is needed for long enough support.
See recommended ruby from https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html

This means Rails 7.2, ruby 3.4 and Ubuntu 24.04 LTS

Currently morph is on 
* Rails version 5.2.8.1 - Security support ended 3 years ago
* ruby 2.7.6 - support ended 2.5 years ago
* Ubuntu Xenial (16.04 LTS) - Security support ended 4.5 years ago
  * nodejs standard package
* mysql-server:5.7.33 
* elasticsearch:7.17.7 - recent enough
* redis:3.2

We should upgrade in steps to:

* https://github.com/openaustralia/morph/pull/1366

* Rails 5.2 → 6.0 
  * Zeitwerk autoloading (replaces classic autoloader)
  * Parallel testing, multi-DB support
  * https://github.com/openaustralia/morph/pull/1367

* Rails 6.0 → 6.1
  * Horizontal sharding, delegated types
  * Strict loading for N+1 prevention
  * https://github.com/openaustralia/morph/pull/1368

* Rails 6.1 → 7.0, Ruby -> 3.0.7
  * Asset pipeline changes (removes Sprockets requirement)
  * Encrypted attributes, query method improvements
  * Check ActiveAdmin compatibility (often lags Rails versions)
  * Includes upgrade to ruby-3.0.7 to resolve gem requirements (Installed on current production server)
  * https://github.com/openaustralia/morph/pull/1369

* Rails 7.0 → 7.1
  * Async queries, composite primary keys
  * Dockerfile generation, authentication generator

*Rails 7.1 security support has just ended (01 Oct 2025), ruby 3.0.7 support ended over a year ago (23 Apr 2024)*

**Production server has to be rebuilt for below**

* Ubuntu 16.04 → 24.04 LTS (Noble) (latest LTS, Support till  25 Apr 2029) - Gets MySQL 8.0, Redis 7.0, modern system packages (Deferred)
[...]

PR #1489, opened 2026-07-29
Apply security updates that are non breaking

ianheggie-oaf:
## Description

Bumps a batch of direct and transitive gem dependencies to their latest versions still compatible with Ruby 2.7.6, closing out the security advisories that `bundle-audit` flagged and don't require a major version bump or a Ruby upgrade first.

Gems updated: RedCloth, sinatra, jwt, dalli, nokogiri (test group), loofah, rails-html-sanitizer, crass, addressable, concurrent-ruby, bcrypt, msgpack, webrick, websocket-driver, oauth2, rexml, rdoc, yard, net-imap, omniauth, rack (transitive floor).

`faraday` and the direct `nokogiri` floor were deliberately left out of this batch — their remaining CVE fixes require Ruby >= 3.0, so they're tracked in a separate follow-up issue alongside the Rails, Devise, Sidekiq and ActiveAdmin upgrades that also need more than a version bump.

## Motivation and Context

`bundle-audit` was reporting a long list of high and medium severity CVEs against our current gem versions. This clears the ones that can be fixed with a same-branch version bump. Follow-up internal issue [#316](https://github.com/openaustralia/oaf-internal/issues/316) tracks the remainder, which need a Ruby upgrade, a Rails upgrade, or other compatibility work first:
* https://github.com/openaustralia/oaf-internal/issues/316

See also:
* https://github.com/openaustralia/morph/issues/1351

## How Has This Been Tested?

Developed and tested on Ubuntu 24.04.4 LTS x86_64 with ruby installed by mise, mysql 8.0, redis 5:7.0, make 4.3.

- [x] Checked affected area manually on my own / staging system
- [x] Ran automated tests on my own system
- [x] Confirmed it passed the GitHub actions tests

## Screenshots (if appropriate):

N/A — dependency version bumps only, no UI changes.

## Types of Changes

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

---

This PR description and the accompanying Gemfile diff were drafted with the assistance of Claude (Anthropic), based on `bundle-audit` output and a review of upstream changelogs to confirm Ruby version compatibility for each gem. All changes were reviewed and tested by a human before merging.

Signed-off-by: Ian Heggie (OAF) <ian@oaf.org.au>
Assisted-by: Claude:claude-sonnet-5


ianheggie-oaf:
@SocketSecurity ignore gem/tilt@2.8.0

The [gem homepage](https://github.com/jeremyevans/tilt) includes a README.md and the gem is well known for running template engines.

(The gemspec uses a trimmed file list rather than the more common use of `git ls-files`)

ianheggie-oaf:
@benrfairless - ready for your review

default_files/ruby/template/Gemfile
```diff
@@ -6,10 +8,10 @@ source "https://rubygems.org"
 
 ruby "3.2.2"
 
-gem "mechanize", "~> 2.8.5"
-gem "nokogiri", "~> 1.19.3"
+gem "mechanize", "~> 2.14.0"
+gem "nokogiri", "~> 1.17.2"
```
benrfairless, line 12: This is a 16 month downgrade and it reintroduces 16 nokogiri advisories that `~> 1.19.3` doesn't have, including GHSA-mrxw-mxhj-p664 (7.8, packaged libxslt CVEs, fixed 1.18.4), GHSA-c4rq-3m3g-8wgx (7.5, CSS selector ReDoS, fixed 1.19.3) and three vendored libxml2 roll-ups. In a PR called "Apply security updates" that needs to be a deliberate, stated trade-off rather than a commit body of "latest gem version that run on heroku-18".

I think I follow the reasoning: 1.17.2 is the last release shipping `x86_64-linux`, and 1.18.0 replaced it with `x86_64-linux-gnu`/`-musl` which target a newer glibc than heroku-18 provides. The `platform` file here is indeed `heroku-18`, so the goal is coherent. But see the sqlite3 comment below, because the same commit moves that gem the other way across the same boundary.

Also note the template has been shipping post-split `nokogiri 1.19.3-x86_64-linux-gnu` to `heroku-18` since #1475. If that's genuinely broken we'd expect scraper failures; if it isn't, this downgrade isn't buying anything.

One option worth considering instead: move this template to `heroku-24` and keep nokogiri 1.19.x. `default_files/python/template/platform` is already on heroku-24, so there's precedent.

Whichever way it goes, `default_files/ruby/template/Gemfile.lock` needs regenerating to match - right now it still declares `nokogiri (~> 1.19.3)` and this Gemfile says `~> 1.17.2`.

default_files/ruby/template/Gemfile
```diff
@@ -6,10 +8,10 @@ source "https://rubygems.org"
...
 gem "scraperwiki", git: "https://github.com/openaustralia/scraperwiki-ruby.git", branch: "morph_defaults"
-gem "sqlite3", "~> 1.6.3"
+gem "sqlite3", "~> 2.2.0"
```
benrfairless, line 14: This is the bit that makes me unsure the heroku-18 rationale holds. sqlite3 crosses the same platform split as nokogiri, but in the opposite direction: 1.6.3 ships an `x86_64-linux` native gem, 2.2.0 ships only `x86_64-linux-gnu` and `-musl`.

So either the glibc theory is right, in which case this bump breaks heroku-18 and the commit doesn't achieve its stated goal, or both gems fall back to the plain `ruby` platform gem and compile from source, in which case nokogiri 1.19.x would too and the downgrade above is unnecessary.

Could you confirm which it is with a build on heroku-18? If it's the second, `BUNDLE_FORCE_RUBY_PLATFORM` would let the template keep nokogiri 1.19.x and we'd avoid the security regression entirely.

Gemfile
```diff
@@ -70,15 +70,15 @@ gem "multiblock"
...
 gem "rails-timeago", "~> 2.0"
 # For sidekiq ui
-gem "sinatra", ">= 1.3.0", require: nil
+gem "sinatra", ">= 2.2.3", require: nil # fix CVE-2022-45442; CVE-2024-21510 and CVE-2025-61921 need sinatra 4.x, tracked separately
 gem "sitemap_generator"
 gem "zeroclipboard-rails"
 
 # For the administration interface
-gem "activeadmin"
+gem "activeadmin", ">= 2.12.0" # fix CVE-2023-50448; CVE-2023-51763 and CVE-2024-37031 need >= 3.2.0, which needs Rails 6.1+, tracked separately
```
benrfairless, line 78: The comment says 3.2.0 needs Rails 6.1+, but the constraint doesn't encode that, so the next `bundle update` will happily try to resolve ActiveAdmin 3.x or 4.x against Rails 6.0.6.1. `rack` a few lines down gets this right with `">= 2.2.23", "< 3"` - suggest matching that shape here:

```suggestion
gem "activeadmin", ">= 2.12.0", "< 3" # fix CVE-2023-50448; CVE-2023-51763 and CVE-2024-37031 need >= 3.2.0, which needs Rails 6.1+, tracked separately
```

Same thought applies to `puma` and `sinatra` above if staying on the current major was the intent - both jumped a major on this PR (5.6.9 → 6.6.1 and 2.2.4 → 3.2.0) because the floors are unbounded.

benrfairless (commented):
Checked every advisory claim against a fresh clone of `rubysec/ruby-advisory-db` and every version against the rubygems.org API. All 24 CVE/GHSA IDs in the comments are real and correctly attributed, which was the first thing worth confirming given the LLM assistance note. For the app itself this does what it says: 72 unpatched advisories on `main` down to 67 here, five genuinely cleared, none introduced.

Two things look blocking to me, both in `default_files/ruby/template/`, plus a couple of constraint tweaks inline.

**1. `default_files/ruby/template/Gemfile.lock` wasn't regenerated.** The Gemfile now asks for `mechanize ~> 2.14.0`, `nokogiri ~> 1.17.2`, `sqlite3 ~> 2.2.0`, but the lock beside it still declares `~> 2.8.5`, `~> 1.19.3`, `~> 1.6.3` and pins `nokogiri 1.19.3-x86_64-linux-gnu` / `sqlite3 1.6.9-x86_64-linux`. Both files ship into every new Ruby scraper, so a frozen/deployment bundle fails outright and an unfrozen one silently re-resolves. Nothing in CI covers this: `rspec` runs with `DONT_RUN_DOCKER_TESTS=1`, which skips the only specs that build from the template. Worth also checking `PLATFORMS` (still `x86_64-linux`, which sqlite3 2.x no longer publishes) when regenerating.

**2. The nokogiri downgrade is a security regression** and its rationale conflicts with the sqlite3 bump in the same commit. Details inline.

Smaller stuff, not blocking:

- The bare `>=` floors let four majors through: puma 5→6, sinatra 2→3, jwt 2→3, ransack 2→3 (via ActiveAdmin). Worth a sentence in the description since the title says non-breaking. `/admin/jobs`, the GitHub App JWT path and the ActiveAdmin filters have no spec coverage, so they're worth a manual smoke test. I did check jwt 3.x and Sidekiq::Web separately and both look fine: the 3.0.0 breaking changes (HS512256, deprecated claim verification, rbnacl) don't touch `JWT.encode(payload, rsa_key, "RS256")`, and rack-protection 3.2.0 still has `AuthenticityToken` with the same `use:` option Sidekiq 5.2.10 passes.
- The description says ActiveAdmin is deferred to #316, but 371fb4fb bumps it to 2.12.0. It also doesn't mention puma, the template changes or the Gemfile sort. And of the 22 floors added only 5 advisories actually cleared (activeadmin CVE-2023-50448, addressable CVE-2026-35611, jwt CVE-2026-45363, rails-html-sanitizer CVE-2026-73648, loofah CVE-2026-73490) - the rest were already satisfied, so they're ratchets rather than updates. Fine thing to do, just a different claim, and this description is the durable record of why the pins exist.
- Suggestion for a follow-up rather than here: add `bundle-audit check --update` to `ruby.yml`, otherwise these hand-maintained comments drift and the next regression gets found the same manual way. Also worth noting for #316 that with ActiveStorage disabled (`config/application.rb:9`) the top live residual is omniauth CVE-2015-9284, and `OmniAuth.config.allowed_request_methods = [:post]` mitigates it on 1.9.2 without the blocked 2.0 upgrade.

The honest "what this doesn't fix and why" comments are genuinely useful, thanks for writing those down.

Assisted-by: Claude Code:claude-opus-5

Gemfile
```diff
@@ -104,10 +104,10 @@ gem "sorbet-static-and-runtime"
 gem "psych", ">= 3.3.4", "< 4"
 
 # For making JSON Web Tokens used by Github API
-gem "jwt"
+gem "jwt", ">= 2.10.3" # fix CVE-2026-45363
```
benrfairless, line 107: This floor doesn't actually guarantee a patched version. CVE-2026-45363 is patched in `~> 2.10.3` **or** `>= 3.2.0`, so everything from 2.11.0 to 3.1.2 satisfies `>= 2.10.3` and is still vulnerable. The lock happens to land on 3.2.0 so there's no live exposure, but the constraint doesn't say what the comment says.

```suggestion
gem "jwt", ">= 3.2.0" # fix CVE-2026-45363 (patched in ~> 2.10.3 or >= 3.2.0, so 2.11.0-3.1.2 are still affected)
```

For what it's worth I checked the jwt 3.x upgrade itself and it's safe here - the 3.0.0 breaking changes are dropping HS512256, removing deprecated claim verification methods and dropping rbnacl, none of which touch `JWT.encode(payload, private_key, "RS256")` in `github_app_installation.rb:236`.

benrfairless (approved):

PR #1492, opened 2026-08-01
Chore/first time booting papercuts with docker-up

mathewdbutton:
## Description

These are the errors I faced in attempting to get the app running with `make docker-up`, and the commits that went along to fix them. I've split them into far too granular pieces but it was to aid in figuring out what was relevant and what should be thrown away.

  1. MySQL container failing healthcheck
  Error: Access denied for user 'root'@'127.0.0.1' (using password: YES) or (using password: NO)
  Root cause: healthcheck wasn't using password authentication correctly

  2. Char set error - split out to this [PR](https://github.com/openaustralia/morph/pull/1493)

  3. Web and worker containers can't connect to Redis/MySQL
  Error: Error connecting to Redis on redis:6379 (SocketError) and similar for MySQL
  Root cause: Ruby services weren't on the same Docker network as the backing services

  4. Model annotations showing incorrect defaults
  Error: Annotate gem removing default("") from string columns and duplicating foreign key comments
  Root cause: Charset change caused annotate to re-read schema metadata differently

## Motivation and Context

Hello, first time contributor (hopefully), Ben Fairless put the call out for help in the Ruby Oceania slack workspace. This was my first run trying to get everything working purely inside of docker, using `make docker-up`.

I've tried to follow the contributor guidelines but I do feel like I've missed some, my apologies in advance.

## How Has This Been Tested?

- [x] Checked affected area manually on my own / staging system
I was able to successfully boot up the app with `make docker-up` and click around
- [ ] Ran automated tests on my own system
- [ ] Confirmed it passed the GitHub actions tests

## Types of Changes

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

I have read the CLA Document and I hereby sign the CLA

### DIFF
--- a/spec/models/scraper_spec.rb
+++ b/spec/models/scraper_spec.rb
@@ -35,6 +35,8 @@
 # Foreign Keys
 #
 #  fk_rails_...  (create_scraper_progress_id => create_scraper_progresses.id)
+#
+#  fk_rails_...  (create_scraper_progress_id => create_scraper_progresses.id)
 
 require "spec_helper"
 
--- a/app/models/domain.rb
+++ b/app/models/domain.rb
@@ -9,7 +9,7 @@
 #
 #  id         :integer          not null, primary key
 #  meta       :text(65535)
-#  name       :string(255)      default(""), not null
+#  name       :string(255)      not null
 #  title      :text(65535)
 #  created_at :datetime
 #  updated_at :datetime
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,6 +26,8 @@ x-ruby-service: &ruby-service
     MYSQL_USER: "root"
     REDIS_URL: "redis://redis:6379"
     MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PASSWORD:?MYSQL_ROOT_PASSWORD must be set, preferably in the .env file}
+  networks:
+    - network
 
 services:
   web:
--- a/docker_images/services.yaml
+++ b/docker_images/services.yaml
@@ -43,7 +43,7 @@ services:
       MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PASSWORD:-testrootpass}
       MYSQL_ROOT_HOST: '%'
     healthcheck:
-      test: [ "CMD", "mysqladmin", "ping", "-h", "localhost" ]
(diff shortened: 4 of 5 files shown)


spec/models/scraper_spec.rb
```diff
@@ -35,6 +35,8 @@
 # Foreign Keys
 #
 #  fk_rails_...  (create_scraper_progress_id => create_scraper_progresses.id)
+#
+#  fk_rails_...  (create_scraper_progress_id => create_scraper_progresses.id)
```
mathewdbutton, line 39: I think the annotation gem might be having a bit of trouble here, not sure what was going on but left it here so others didn't get the same problem. It could be due to the age of the version being used.

CloCkWeRX: Oh good; I wasn't the only one that ran into this!

CloCkWeRX:
For 3bc35c1710d4c70bd9e46fa3db9f7d73d76a74ef , maybe that's worth doing standalone.

Depending on what's on staging/prod; will need to do basically https://docs.cloudbees.com/docs/cloudbees-cd-kb/latest/cloudbees-cd-kb/kbec-00515-converting-a-mysql-database-from-utf8-utf8mb3-to-utf8mb4 which would need a bit of a manual fussing about at deployment time.

mathewdbutton:
> For https://github.com/openaustralia/morph/commit/3bc35c1710d4c70bd9e46fa3db9f7d73d76a74ef , maybe that's worth doing standalone.

Yeah agreed, it's the only one that looks slightly hairy. Oh that's a great article you've linked to, thanks. I really wasn't sure which way to go with this, but it seems like it's work either way.

> The following steps show how to create a DB dump, edit the DB dump so that the correct charset and collation are used, and then restore the new DB.

:fearful: - Oh neat, fussing indeed this feels less than ideal.

spec/models/scraper_spec.rb
mathewdbutton: Oh good, I not going crazy either then

mathewdbutton:
I've split out the charset and schema.rb change to https://github.com/openaustralia/morph/pull/1493

ianheggie-oaf (approved):
Approved - makes sense to me

Issue #1506, opened 2026-08-17
Repository cleanup: remove dead code and fix documentation drift


benrfairless:
An end-to-end review of the repository identified dead code and documentation drift. The cleanup is split into five focused PRs, all branching off `main`:

1. **Dead app code** - routes pointing at missing controllers/actions, a fully commented-out ActiveAdmin file, superseded PhantomJS documentation partials, an unrendered statistics partial, five methods with no callers, and a rake task broken since 2015.
2. **Dead assets** - empty scaffold CoffeeScript/SCSS files and unreferenced images.
3. **Unused gems** - `multiblock` and top-level `ruby-progressbar`.
4. **README/TESTING alignment** - remove contribution/branch-naming sections that duplicate (and contradict) the inherited org-level [CONTRIBUTING.md](https://github.com/openaustralia/.github/blob/main/.github/CONTRIBUTING.md), replace the drifted hand-copied make targets list with a pointer to `make help`, fix a broken link to a nonexistent doc, consolidate test-running instructions into TESTING.md, and fix typos.
5. **Support file drift** - duplicated section in `provisioning/README.md`, duplicated/stale content in `env-example`, a mangled comment in `docker-compose.override.yml-example`, `.PHONY` typos in the Makefile, and commented-out blocks in `.github/workflows/ruby.yml`.

Deliberately out of scope (may warrant future issues):

- Dropping unused DB columns (`scrapers.forked_by_id`, `scrapers.scraperwiki_url`, `owners.feature_switches`) - needs migrations
- The mitmproxy/connection-log ingestion path, which is disabled but still has display/maintenance code ("zombie" state) - needs a product decision
- Medium-confidence gem pruning (`uglifier`, `sdoc`, `growl`) and duplicate rake tasks in `lib/tasks/docker.rake`
- `.rubocop_todo.yml` burn-down

This review and the resulting PRs were produced with AI assistance (OpenCode with anthropic.claude-fable-5), with human review.

benrfairless:
All five scoped PRs (#1507, #1508, #1509, #1510, #1511) are merged and were deployed to production in the 82fe0fb..ec15812e upgrade on 8 September 2026, with the site confirmed running ec15812.

The items the issue listed as deliberately out of scope remain so, and are not tracked by this issue: dropping unused DB columns, the disabled mitmproxy/connection-log path (needs a product decision), medium-confidence gem pruning, and the .rubocop_todo.yml burn-down. They can get their own issues if and when someone picks them up.

[...]

PR #1508, opened 2026-08-17
Remove empty scaffold assets and unused images

benrfairless:
Part of #1506.

## What

- Deletes seven generator-created `.js.coffee` files that contain only the three-line scaffold comment. Three of them (`api`, `documentation`, `users`) were still `//= require`d in `application.js`, so those lines go too; the other four were never required at all.
- Deletes two empty scaffold stylesheets (`connection_logs.scss`, `site_settings.scss`) - `application.scss` imports files explicitly and there is no `require_tree`, so they were never included.
- Deletes three unreferenced images under `app/assets/images/examples/`.

## How verified

- Grepped every deleted filename/image path across `app/`, `lib/`, `config/`, `public/` (views, SCSS, helpers, JS) - zero references
- Confirmed no `require_tree` directive and nothing in `config/initializers/assets.rb` precompiles the deleted files
- Asset compilation exercised by CI (sprockets loads on boot for the test suite); will take out of draft once green

## AI assistance

Prepared with AI assistance (OpenCode with anthropic.claude-fable-5); reviewed by a human before submission.


ianheggie-oaf (approved):
Approved, makes sense to me

PR #1511, opened 2026-08-17
Fix drift and duplication in support files

benrfairless:
Part of #1506.

## What

- **`provisioning/README.md`** - the entire "Production devops development" section was duplicated verbatim back-to-back (copy-paste/merge artifact). Merged into one copy that keeps the superset of both: the `vagrant-vbguest` plugin (first copy only) and the `cap local deploy` tail (second copy only). Also fixes `make ansible` → `make production-provision` (the `ansible` target doesn't exist).
- **`env-example`** - three blocks each appeared twice: the `STAGING_HOSTNAME`/`STAGING_DEPLOY_TO` settings, the "Code Development Only" header, and the elasticsearch guidance. Merged each. Also removed guidance referencing the long-dead `docker-machine` and old `docker-compose up` syntax, commented out `ELASTICSEARCH_URL` (`make services-up` exposes elasticsearch on `localhost:9200`, which is also the searchkick default - the previous uncommented `192.168.99.100` was a docker-machine IP that broke anyone copying the file verbatim), and dropped the pointer to OS X docker setup instructions that no longer exist in the README.
- **`docker-compose.override.yml-example`** - the first line said "copy  to docker-compose.override.yml" with the source filename missing (botched edit).
- **`Makefile`** - `.PHONY` listed four names that are not targets: `deploy`, `provision`, `up`, and the misspelt `staging-anible`.
- **`.github/workflows/ruby.yml`** - removed the commented-out mysql `services:` block and the `stack-version` remnant.

## How verified

- `make help` renders the full target list correctly after the `.PHONY` change
- Workflow YAML parses (`YAML.load_file`)
- Confirmed `docker_images/services.yaml` maps elasticsearch to host port 9200 before changing the `ELASTICSEARCH_URL` example
- Diffed the two duplicated provisioning sections to confirm the merged copy loses nothing

## AI assistance

Prepared with AI assistance (OpenCode with anthropic.claude-fable-5); reviewed by a human before submission.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -39,16 +39,6 @@ jobs:
       DB_DATABASE: scraping_test
       DB_USER: scraping
       DB_PASSWORD: password
-    # services:
-    #   mysql:
-    #     image: mysql:8.0
-    #     env:
-    #       MYSQL_DATABASE: scraping_test
-    #       MYSQL_USER: scraping
-    #       MYSQL_ROOT_PASSWORD: password
-    #     ports:
-    #     - 33306:3306
-    #     options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
     steps:
       - name: Checkout this repo
         uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


ianheggie-oaf (approved):
Approved - lgtm at a quick read through,  thanks for the cleanup

PR #1517, opened 2026-08-21
Only follow safe referers after sign out

### LINKED ISSUE #1438: Sign out redirects to unsafe client supplied request.referer
**Description**

Morph redirects to client provided request.referer on sign_out which is unsafe - for example read [Unmasking the Danger: Open Redirection via Referer Header](https://medium.com/@sulmanfarooq531/unmasking-the-danger-open-redirection-via-referer-header-a527ab4e7be1)

**To Reproduce**
Steps to reproduce the behaviour:
1. Attacker creates page on https://evil.example.com with logout link to your site
2. User legitimately clicks on link to https://morph.io/sign_out, or is redirected through the evil site from a link in an email, txt etc
3. Browser sets Referer: https://evil.example.com/signed_out
4. morph.io redirects them back to https://evil.example.com/signed_out
5. Evil site shows a fake login page as if its morph.io / github
6. User attempts to login thinking they are on morph / github

**Expected behavior**

We either:
1. Always redirect to the root page ("/")
2. We validate the given referrer details and only redirect to it if 
  * Its our domain
  * Its a page that can be displayed without authentication

**Additional context**

Found when looking at test coverage. `ApplicationController#after_sign_out_path_for` has no coverage.

```ruby
  sig { params(_resource: T.untyped).returns(T.nilable(String)) }
  def after_sign_out_path_for(_resource)
    request.referer ? URI.parse(request.referer).path : root_path
  end
```

Alternative solution

```ruby
def after_sign_out_path_for(_resource)
  return root_path unless request.referer.present?
  
  referrer_uri = URI.parse(request.referer)
  
  # Only redirect within your own domain with no query or fragment present
  return root_path if referrer_uri.host != request.host || referrer_uri.query || referrer_uri.fragment
  
  path = referrer_uri.path 
  # Only allow simple paths: /something or /something/else
  return root_path unless path&.match?(%r{\A(/[^/?#]+){1,2}\z})
  
  path
rescue URI::InvalidURIError, TypeError
  root_path
end
```

Note identical named method `Devise::Controllers::Helpers#after_sign_out_path_for` is overly complicated - could be simplified to just returning `root_path`

benrfairless:
Fixes #1438.

## What this does

`ApplicationController#after_sign_out_path_for` followed the client-supplied `request.referer` blindly, which is an open redirect: an attacker can bounce a user through morph.io's sign-out and back to a phishing page (see the reproduction steps in #1438). Returning only the parsed *path* did not close the hole either, because a referer like `https://morph.io//evil.example.com/x` yields the protocol-relative path `//evil.example.com/x`, which browsers treat as a full URL on the other host. A malformed referer also raised `URI::InvalidURIError` (a 500).

This implements the issue's option 2 (validated referer): only redirect back to the referer when it is on our own host, has no query or fragment, and is a simple one- or two-segment path like `/some_user/some_scraper`; everything else falls back to the root path.

One deliberate deviation from the sketch in the issue: the `TypeError` rescue is omitted, because the `blank?` guard means `URI.parse` only ever receives a non-empty string, which cannot raise `TypeError`.

One part of option 2 is deliberately not enforced: "a page that can be displayed without authentication". The path regex admits auth-required pages like `/settings`, matching the issue's own proposed code, which has the same property. The worst case there is a bounce to the sign-in redirect on our own host, not an open redirect, and maintaining a whitelist of unauthenticated routes in the controller did not seem worth that trade.

## Testing

TDD'd: the rewritten specs (cross-host referer, doubled slash, query string, fragment, deep path, relative referer, invalid URI) were run red against the old implementation first - six of them failed, confirming they detect the vulnerability - then green against the fix.

- Target spec: 12 examples, 0 failures
- Full CI-profile suite (`DONT_RUN_DOCKER_TESTS=1 RUN_SLOW_TESTS=1 DONT_RUN_GITHUB_TESTS=1`) in the Docker dev environment: 848 examples, 0 failures, 17 pending, coverage 81.39%
- RuboCop clean; `srb tc` unchanged against the pre-existing baseline

Note on coverage: the quick-tests SimpleCov threshold (77.08) fails in my container on `main` as well (76.67 baseline vs 76.72 with this change - coverage goes up, not down), so the hardcoded threshold has not been touched; CI's own profile is the gate that counts.

## AI disclosure

This change was written with the assistance of an AI coding agent (OpenCode running anthropic.claude-fable-5), driven and reviewed by @benrfairless.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,10 @@
 class ApplicationController < ActionController::Base
   extend T::Sig
 
+  # Only redirect back to simple one- or two-segment paths like "/documentation"
+  # or "/some_user/some_scraper" after sign out
+  SAFE_SIGN_OUT_PATH_REGEXP = T.let(%r{\A(/[^/?#]+){1,2}\z}.freeze, Regexp)
+
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
@@ -30,9 +34,22 @@ class ApplicationController < ActionController::Base
     session[:previous_url] = request.fullpath unless request.fullpath =~ %r{/users}
   end
 
-  sig { params(_resource: T.untyped).returns(T.nilable(String)) }
+  # The referer is client supplied, so only follow it back to a simple path on
+  # our own host. Anything else is an open redirect (see issue #1438).
+  sig { params(_resource: T.untyped).returns(String) }
   def after_sign_out_path_for(_resource)
-    request.referer ? URI.parse(request.referer).path : root_path
+    referer = request.referer
+    return root_path if referer.blank?
+
+    uri = URI.parse(referer)
+    return root_path if uri.host != request.host || uri.query || uri.fragment
+
+    path = uri.path
+    return root_path if path.nil? || !path.match?(SAFE_SIGN_OUT_PATH_REGEXP)
+
+    path
+  rescue URI::InvalidURIError
+    root_path
   end
 
   # Overriding the default ability class name used because we've split them out. See
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -17,24 +17,72 @@ describe ApplicationController do
   end
 
   describe "#after_sign_out_path_for" do
+    def stub_referer(referer)
+      allow(controller).to receive(:request).and_return(
+        instance_double(ActionDispatch::Request, referer: referer, host: "morph.io")
+      )
+    end
+
+    def after_sign_out_path
+      controller.send(:after_sign_out_path_for, :any_scope)
+    end
+
     before do
       allow(controller).to receive(:root_path).and_return(root_path)
     end
 
-    # FIXME: This is not safe to do!
-    it "returns referer path" do
-      referer = "/some_user/some_scraper"
-      allow(controller).to receive(:request).and_return(instance_double(ActionDispatch::Request, referer: referer))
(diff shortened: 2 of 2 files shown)


ianheggie-oaf (approved):
Approved - LGTM at a quick read once spec and code

app/controllers/application_controller.rb
```diff
@@ -4,6 +4,10 @@
 class ApplicationController < ActionController::Base
   extend T::Sig
 
+  # Only redirect back to simple one- or two-segment paths like "/documentation"
+  # or "/some_user/some_scraper" after sign out
+  SAFE_SIGN_OUT_PATH_REGEXP = T.let(%r{\A(/[^/?#]+){1,2}\z}.freeze, Regexp)
```
CloCkWeRX, line 9: Is it better to replace this with a known good list from config/routes.rb?

To avoid reflection attacks/DoS could just setup a GitHub action to bundle exec rake routes and allowlist the results in code every 6 months.

(I mean, go with this for now but consider explicit allowlists for later)

benrfairless: The regex is backtracking-safe — bounded {1,2} repetition and /-delimited segments that exclude / from the class, so it's linear time and not ReDoS-able. On the allowlist idea: our main redirect target is the wildcard route /:owner/:scraper, so a list generated from rake routes would still have to admit arbitrary two-segment paths — same acceptance set as the regex, plus a CI artifact that can drift. If we ever want real route validation, Rails.application.routes.recognize_path checks the live routing table at runtime with no generated list. Happy to open a follow-up issue if you think it's worth tracking.

Issue #1474, opened 2026-05-07
Add Sentry.io to replace Honeybadger


benrfairless:
**Is your feature request related to a problem? Please describe.**
We constantly hit monthly limits with Honeybadger, and they no longer offer a OSS Plan

**Describe the solution you'd like**
Sentry.io has given us a sponsored Business Plan. We should take advantage of this.

**Describe alternatives you've considered**
Keep Honeybadger and deal with hitting limits, or add NewRelic (which doesn't have as many features and is harder to configure)

**Additional context**
Please speak with me for access to Sentry before proceeding.

If possible we should run both Honeybadger and Sentry for a period to verify no issues.

ianheggie-oaf:
@benrfairless please confirm that this is focused on monitoring errors, not replacing NewRelic server monitoring?

Also cronjob monitoring is currently the emails we get from cron - we could add sentry monitoring, depending if the sponsored bus plan includes more than one cron monitor or you are find with the extra we would need.

I suggest we have a separate issue to track monitoring cron jobs (we are missing noticing when some fail).

Also APM monitoring is not currently set up - does the sentry plan include sufficient allowance?

(cc @Br3nda )

PR #1518, opened 2026-08-21
Add Sentry alongside Honeybadger

benrfairless:
Advances #1474, and completes Phase 1 item 2 of [`doc/UPGRADE_PATH.md`](https://github.com/openaustralia/morph/blob/main/doc/UPGRADE_PATH.md): with no staging environment, Sentry is the soak instrument for every Rails hop that follows, so it lands before the framework work starts.

## What this does

Adds `sentry-ruby` and `sentry-rails` per the plan, plus `sentry-sidekiq` so background jobs are covered as Honeybadger already covers them today. All resolve to 6.7.0, the current line supporting Ruby 2.7.

Without `SENTRY_DSN` set the SDK initialises but sends nothing, so development and test stay quiet by default. Events are tagged with `APP_VERSION` as the release, with Rails and HTTP breadcrumbs enabled. Honeybadger stays until Sentry has proven itself; removal is planned for Phase 8, per the issue's own request to "run both Honeybadger and Sentry for a period to verify no issues".

Most of the diff is `bundle exec tapioca gem` output: the checked-in RBIs had not been regenerated since earlier gem bumps, so this run catches up on those as well as adding the sentry gems' RBIs. That refresh exposed one real gap: the stale ransack 2.5.0 RBI had been accidentally providing `search` on ActiveRecord models, masking that searchkick's dynamic `search` had no RBI. Hand-written shims for `Owner.search` and `Scraper.search` (matching the existing `reindex` shim in `sorbet/rbi/scraper.rbi`) close that.

## Testing

TDD'd: the new spec was run red first (`uninitialized constant Sentry`), then green after the gems and initializer were added.

- Target spec: 4 examples, 0 failures
- Full CI-profile suite (`DONT_RUN_DOCKER_TESTS=1 RUN_SLOW_TESTS=1 DONT_RUN_GITHUB_TESTS=1`) in the Docker dev environment: 845 examples, 0 failures, 17 pending, coverage 81.35% (threshold 80.05, untouched)
- RuboCop clean on the changed Ruby files; `srb tc` application-code errors identical to `main` (one net new error exists inside tapioca-generated RBIs, the same unresolved-cross-reference noise as the 14 already on `main`)
- Verified end to end against the production Sentry project: a test event sent through this initializer with the real DSN arrived as [MORPHIO-1](https://oaf-org-au.sentry.io/issues/MORPHIO-1), since resolved. Delivery needs a real DSN so it is recorded as a manual test in `TESTING.md`.

## Deployment note

`SENTRY_DSN` needs to be added to the server's `.env` at deploy time (same treatment as `HONEYBADGER_API_KEY`). The value is the "Production" client key on the `oaf-org-au/morphio` Sentry project.

## AI disclosure

This change was written with the assistance of an AI coding agent (OpenCode running anthropic.claude-fable-5), driven and reviewed by @benrfairless.


ianheggie-oaf (approved):
Approved - LGTM at a glance

PR #1522, opened 2026-08-24
Support development and deploys on Apple silicon

benrfairless:
## What this does

Makes the containerised dev environment the supported route for developing morph on macOS (Apple silicon), and lets Capistrano deployments run from inside the container. A new devcontainer wraps the existing compose files (one definition of the Ruby environment, no drift) and forwards the host's SSH agent, so `bundle exec cap staging deploy` works from a devcontainer shell without copying keys into the container.

[ADR 0006](docs/adr/0006-containerised-dev-environment-via-devcontainer.md) records why native macOS development stays unsupported: `Gemfile.lock` locks only `x86_64-linux`, and the locked `sorbet-static` ships Darwin builds only up to `universal-darwin-22`, so bundler cannot resolve the bundle on any current macOS at all. Revisit after the Ruby 3.4 move (ADR 0005).

## Pre-existing bugs fixed along the way

Nobody could previously complete the test suite inside the container, and doing so surfaced these:

- A db backup spec calls code that `abort`s when `mysqldump`/`zstd` are missing from the image, killing RSpec mid-suite with a seed-dependent truncated summary and 0 reported failures. The image now includes both, plus the docker CLI so `spec_helper.rb`'s `docker info` detection can enable the `:docker` specs.
- `ELASTICSEARCH_URL` in `docker-compose.yml` pointed at port 9288 instead of 9200.
- `Morph::DockerUtils` pulls and builds now request `linux/amd64` explicitly. The scraper images only exist for amd64, and an ARM64 daemon otherwise refuses them ("no matching manifest for linux/arm64/v8"). No behaviour change on production, which is already amd64.
- Rosetta writes its AOT cache to `$HOME/.cache` inside containers, and herokuish builds with `HOME=/app`, so the build's final `mv /tmp/build/.cache /app/.cache` can race a freshly recreated `/app/.cache` and fail an otherwise successful build. `Morph::DockerRunner.compile` recovers from exactly that case (guarded so a genuine compile failure still fails).

## Other changes

- `mitmdump2` moves behind a `proxy` compose profile and out of `web`'s `depends_on`: the proxy path it serves is switched off in production (#1506) and the image is amd64-only. No product code touched.
- Corrected the copy-pasted postgis comment in `docker_images/services.yaml` (the real reason for the amd64 pin is that `mysql:5.7` has no arm64 build).
- New make targets: `devcontainer-up` / `devcontainer-shell`, and `dev-scraper-images` / `dev-scraper-network` for the host-side setup scraper runs need on Docker Desktop's containerd image store (details in the ADR). `docker-up` is out of BETA.
- The `quick-tests` coverage threshold is recalibrated from 77.09 to the reproducible in-container figure of 76.67 (native Linux runs measure slightly higher, matching the existing FIXME about CI-vs-local drift).

## Verified in the containerised environment

- Rails boots, DB migrates, and `make quick-tests` passes: 829 examples, 0 failures, exit 0 (previously truncated to ~400 seed-dependent examples by the abort above)
- Scraper end-to-end: nodejs-on-heroku-18 and python-on-heroku-24 `:docker`/`:slow` specs compile and run real scraper containers
- `cap` loads with ssh and the deploy gems present in the image
- rubocop clean on changed Ruby files

## Known limits

- The ruby/php/perl heroku-18 examples in `docker_runner_spec` fail from pre-existing fixture rot (the example scraper's Gemfile and Gemfile.lock are out of sync, so bundler's deployment mode refuses). That fails on native Linux too and deserves its own issue.
- Rare containerd-store export flakes on legacy-builder builds remain possible on Docker Desktop; disabling "Use containerd for pulling and storing images" is a machine-level workaround, deliberately not required.

## AI disclosure

This change was developed with the assistance of an AI coding agent (Claude, model anthropic.claude-fable-5), including the investigation, code, docs and this PR description. A human reviewed and directed the work throughout.


ianheggie-oaf (approved):
Approved - excellent work to get it going for yourself. I am curious how fast or slow the emulation is for you.

PR #1526, opened 2026-09-09
Fix ansible playbook check-mode failures

benrfairless:
## What this does

Running the playbook against production with `--check --diff` failed part way through. This gets the full production check to complete with `failed=0`, and fixes one latent bug that would have bitten a real run.

Three commits, separable if needed:

1. **Make ansible playbook runnable in check mode.** Read-only or localhost-only tasks (swap checks, GitHub pubkey downloads, bundler presence check) now set `check_mode: false` so the tasks that consume their results work under `--check`; none of them changes the managed host. `async` values (only there to avoid PTY hangs) are templated away under check mode, because Ansible refuses to combine async with check.

2. **Guard the placeholder symlink so it cannot clobber a live deploy.** `force: no` on the `file` module does not stop it repointing an *existing* symlink, so a real playbook run on a deployed host would have repointed `/var/www/current` from the current release to the empty placeholder, taking the site down. The task now stats `/var/www/current` first and only links when nothing is there, matching its stated intent.

3. **Disable the yarn role while its apt repo is broken on xenial.** Since 2026-01-28 the Yarn apt repository signs only with an Ed25519 key, which gpg 1.4 on Ubuntu 16.04 can neither import nor verify, so the repo is unusable on xenial full stop. Nothing in morph currently uses yarn (assets are sprockets, no `package.json`) and it was never successfully installed on production. The role (added in #1406 for the Rails 6 upgrade) is gated behind `install_yarn`, default off, with a comment explaining why. @ianheggie-oaf FYI: when the Rails asset work actually needs yarn it will have to be a tarball/npm install or wait for the OS upgrade.

## Host-side change, done separately

The first check failure ("Failed to update apt cache") was not a playbook bug: Linode rotated the Longview apt repo signing key in November 2025. The new key from https://apt-longview.linode.com/linode.gpg has been installed on morph.io directly (old key backed up at `/root/linode.gpg.bak-2026-09-08`). Nothing in this PR relates to that, but the check will not pass without it.

## Testing

- `.venv/bin/ansible-playbook --user=root --inventory-file=provisioning/inventory/production provisioning/playbook.yml --check --diff` now completes: `ok=132 changed=22 failed=0`.
- `--syntax-check` passes.
- The `changed=22` is genuine drift on production worth reviewing before a real run: 6 pending package upgrades, new authorized_keys entries, help.morph.io nginx log config, and a gem refresh.

## AI disclosure

Assisted-by: OpenCode:amazon-bedrock/anthropic.claude-fable-5


Br3nda:
this is an existing problem but it wants to remove the new lines in the database config

image

Br3nda (approved):

PR #1498, opened 2026-08-07
Raise on Rails deprecation warnings in tests

benrfairless:
## Description

Changes `config.active_support.deprecation` in the test environment from `:stderr` to `:raise`, so any Rails deprecation warning triggered during the test suite now fails the run instead of scrolling past in stderr.

Part of the staged Rails upgrade stack (stacked on #1497, `upgrade/00-config-cleanup`).

## Motivation and Context

This is the baseline guard for the staged Rails upgrade: each Rails version bump in the stack can only merge once the deprecations it introduces are fixed, because CI will raise on them. Without this, deprecation warnings accumulate silently and the upgrade debt compounds.

## How Has This Been Tested?

- [ ] Checked affected area manually on my own / staging system
- [ ] Ran automated tests on my own system
- [x] Confirmed it passed the GitHub actions tests

The full rspec suite passes with `:raise` enabled, confirming no existing deprecations fire under the current Rails version.

## Screenshots (if appropriate):

Not applicable — test configuration change only.

## Types of Changes

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation

None of the above: test/CI configuration change with no runtime behaviour impact.

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Assisted-by: Claude Code/anthropic.claude-fable-5

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #1497, opened 2026-08-07
Remove dead Active Storage configuration

benrfairless:
## Description

Removes Active Storage configuration that has never had any effect:

- `config.active_storage.service` settings in `config/environments/development.rb`, `production.rb`, and `test.rb`
- `config/storage.yml` (service definitions plus commented-out S3/GCS/Azure examples)

The `active_storage/engine` is deliberately not loaded in `config/application.rb` ("Not used nor in schema.rb"), so none of this configuration was ever read.

## Motivation and Context

Part of the dependency upgrade stack (#1495 → #1496 → this → #1498). Dead configuration adds noise and invites wasted effort during the upcoming Rails upgrade — anyone auditing config would otherwise have to rediscover that Active Storage isn't actually in use. There are no Active Storage tables in `schema.rb` and no references in app code (the only remaining mentions are generated Sorbet RBI shims).

## How Has This Been Tested?

- [ ] Checked affected area manually on my own / staging system
- [ ] Ran automated tests on my own system
- [x] Confirmed it passed the GitHub actions tests

Full CI suite (rspec, lint, Brakeman) is green on this branch. Since the engine is not loaded, removing its configuration cannot change runtime behaviour.

## Screenshots (if appropriate):

N/A — configuration-only change.

## Types of Changes

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation

None of the above: chore/cleanup removing dead configuration, with no functional change.

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Assisted-by: Claude Code/anthropic.claude-fable-5

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #1423, opened 2025-12-09
Upgrade from rails 6.0 to rails 6.1 [WIP Q1 2026]

ianheggie-oaf:
## Relevant issue(s)

* https://github.com/openaustralia/morph/issues/1354

## What does this do?

Upgrades morph to rails 6.1

## Why was this needed?

Rails 6.0 is out of date and has outstanding security issues

## Notes to reviewer (Optional)

Waiting till the following PRs are reviewed and merged:
* https://github.com/openaustralia/morph/pull/1429 - Affects what user sees (scraper history)
* https://github.com/openaustralia/morph/pull/1433 - Helps developer
* https://github.com/openaustralia/morph/pull/1432 - Helps developer

PR #1528, opened 2026-09-11
Update Rails from 6.0.6.1 to 6.1.7.10

benrfairless:
## What and why

Phase 1 Rails hop from doc/UPGRADE_PATH.md: Rails 6.0.6.1 to 6.1.7.10, with the psych "< 4" pin replaced by the security floor the Gemfile TODO planned, the two 6.1 API fixes mined from the stale `chore/rails_61` branch (PR #1423), the `connection_config` deprecation fixed so the #1498 deprecation-raise guard passes, and Sorbet gem RBIs regenerated.
Spring is removed rather than worked around: spring 4.3.0 (the newest release Ruby 2.7 can install) fails deterministically on Rails 6.1, every supported workflow already bypasses it (`DISABLE_SPRING=1` in the containerised env, `bundle exec` in CI and the Makefile), and the fixed spring 4.4+ needs Ruby 3.0, so re-adding it is a Phase 2 candidate at most.

- Part of #1354
- Deferred: the issue's "srb tc clean" exit criterion does not match reality - nothing in CI runs `srb tc` and `main` has 106 errors; this branch reduces that to 100 with none new. Suggest amending the issue rather than fixing 100 pre-existing errors here.
- Deferred: bundle-audit findings all require Rails 7+ fixes, already tracked as deferred in oaf-internal#316.
- Deferred: code-scanning alert 13 (Brakeman BRAKE0120, "Support for Rails 6.1.7.10 ended") is inherent to the phased upgrade: 6.1 is a stepping stone per doc/UPGRADE_PATH.md §5, which waives EOL findings with a phase noted, exactly as the open Rails 6.0.6.1 and Ruby 2.7.6 EOL alerts (2 and 3) are tolerated on main today. It needs a manual dismissal (won't fix) to turn the Brakeman check green; later phases retire it properly.
- Not done: the `:github_integration` vs `:github` spec-tag inconsistency in `spec/spec_helper.rb` is harmless (the specs self-skip) and out of scope.
- The mined branch's Active Storage migrations are deliberately skipped (#1497 removed Active Storage); `chore/rails_61` can be deleted once this merges.

## Type of change

- [x] Chore
- [x] Documentation

## How this was checked

In the containerised dev environment: quick profile 842 examples / 0 failures (coverage 76.77% against the 76.67 threshold), CI profile 854 examples / 0 failures (coverage 81.42% against 80.05), `rubocop` and `haml-lint` clean, `srb tc` at 100 errors versus 106 on `main` with none introduced.
CodeRabbit review of the full diff: 0 findings.

- [x] Ran the automated tests
- [x] Ran `/code-review` (or equivalent): fixed any concerns (and rechecked), linked follow-up issues above, or noted why they are not important above
- [x] Documentation updated, or not needed (AGENTS.md version-pins section)

## Deploy runbook (after merge)

1. Deploy in a low-usage window: `cap production deploy`.
2. Watch Sentry; `cap production deploy:rollback` is the fallback.
3. Hold for 2-3 Sentry-quiet days before deploying the follow-up 6.1-defaults PR; the week-long Phase 1 soak starts after that second deploy.

Assisted-by: OpenCode:anthropic.claude-fable-5


Br3nda (commented):
i tested #1534 and this seems good to do

Br3nda (approved):

Issue #1444, opened 2026-01-04
Update example Ruby scraper.rb to heroku-24 and current language / library versions


ianheggie-oaf:
Update example Ruby scraper.rb to heroku-24 and current language / library versions

Update example so we are not recommending insecure versions, related issue:
* https://github.com/openaustralia/morph/issues/1440

My example scraper works: https://morph.io/ianheggie-oaf/example_ruby_scraper with heroku-18  not heroku-24

Dev branch: feature/update_ruby_example_to_heroku_24

# heroku-24 fails
```
Injecting configuration and compiling...
       -----> Ruby app detected
/tmp/buildpacks/01_buildpack-ruby/bin/compile: line 35:   Failed to download a Ruby executable for bootstrapping!

  This is most likely a temporary internal error. If the problem
  persists, make sure that you are not running a custom or forked
  version of the Heroku Ruby buildpack which may need updating.

  url: https://heroku-buildpack-ruby.s3.us-east-1.amazonaws.com/heroku-24/amd64/ruby-3.1.6.tgz
/tmp/tmp.MnHYxBO7eA/bin/ruby: File name too long
```

# heroku-18 OK
```
Injecting configuration and compiling...
Injecting scraper and running...
Running ruby 3.2.2 on Linux 4fc618c58e77 6.14.3-x86_64-linode168 #1 SMP PREEMPT_DYNAMIC Mon Apr 21 19:47:55 EDT 2025 x86_64 x86_64 x86_64 GNU/Linux
6: Example Domain
```

Issue #1443, opened 2026-01-04
Update example scrapers to use heroku-24 buildstep platform and either get them working or remove them from the published list of supported languages


ianheggie-oaf:
**Description**

With [heroku-24](https://github.com/openaustralia/morph/issues/1440) being enabled, we should update the example scrapers, and either get them working, or comment them out of https://morph.io/documentation

**Describe the solution you'd like**

Decide which languages we wish to support going forth and update those examples,
remove documentation of what we are no longer supporting / or add deprecation notice.

ianheggie-oaf:
@benrfairless - For discussion which languages we wish to support going forth - My suggestion is put an hour max on each and then look at what was problematic

PR #1532, opened 2026-09-11
Default new Ruby scrapers to heroku-24 instead of heroku-18

Br3nda:
## What does this do?

Changes the `platform` file in the Ruby scraper template (used when a new scraper is created via morph.io's "create a scraper" flow) from `heroku-18` to `heroku-24`.

## Why was this needed?

`heroku-18`'s buildstep image hasn't been rebuilt since 2022 - CI build/push for it (and `cedar-14`) was disabled on 2025-05-21 (see `openaustralia/buildstep`'s workflow files, "disabling the older builds so the existing working images don't get disturbed"). `heroku-24` is the only stack still actively rebuilt.

Tested a real Ruby scraper end-to-end against the current `openaustralia/buildstep:heroku-24` image: clean `bundle install`, no buildpack issues.

## Scope - deliberately narrow

- Only affects **brand-new** Ruby scrapers created from this point forward. `default_files/ruby/template/platform` is read once, at creation time, by `CreateScraperWorker` - existing scrapers (with or without their own `platform` file) are completely unaffected.
- `DEFAULT_PLATFORM` (`cedar-14`, used when a scraper has no `platform` file at all) is untouched - so old scrapers relying on that implicit default keep behaving exactly as they do today.
- Ruby only. Node/PHP/Perl are known or suspected broken on `heroku-24` (see the existing `# FIXME: get php and perl examples / buildstep working for heroku-24` in `Language`, and `openaustralia/morph#1443`/`#1444`/`#1448`) and haven't been individually re-verified - happy to do the same test-first pass on each if useful.
- `default_files/python/template/platform` was already `heroku-24` before this PR.

## Testing done

`herokuish buildpack build` against a real scraper (unmodified Gemfile pinning `ruby "3.2.2"`) run directly against `openaustralia/buildstep:heroku-24`, pulled fresh:

```
-----> Ruby app detected
-----> Using Ruby version: ruby-3.2.2
Bundle complete! 3 Gemfile dependencies, 21 gems now installed.
```

Opening as a draft while I sanity-check whether there's anything else worth doing alongside this (e.g. whether `update_docker_images!` needs to run for the change to actually take effect on morph.io's build hosts).


benrfairless (changes requested):
This can't be done while we are on the old server. It will need to wait until we are on the new server with newer docker.

PR #1533, opened 2026-09-11
Make api-triggered runs (morph-cli) honor a scraper's platform file

Br3nda:
## What does this do?

Makes api-triggered runs (`morph execute` / morph-cli) honor the scraper's own `platform` file, instead of always silently building against `DEFAULT_PLATFORM` (`cedar-14`).

## Why was this needed?

`ApiController#run_remote` untars uploaded code straight into `run.repo_path` and creates a `Run` with no `scraper` association at all - there's no git repo involved. Platform selection was `run.scraper&.platform`, which is always `nil` for these runs, so every `morph execute` run silently fell through to `cedar-14` regardless of what the scraper's own `platform` file said.

Found this by triggering a real scraper (`platform: heroku-24`) via morph-cli and getting `cedar-14`'s old `Bundler::LockfileError` instead of a heroku-24 build.

## What changed

- Extracted the platform-file-reading logic already in `Scraper#platform` into `Morph::PlatformFile.read(repo_path)`.
- Added `Run#platform`, which uses the associated scraper's platform when there is one, and otherwise reads the platform file straight out of `run.repo_path` - which already has the uploaded code unpacked into it by this point, including the `platform` file if the scraper has one (morph-cli uploads every file in the directory, this wasn't a client-side gap).
- `Morph::Runner#compile_and_start_run` now calls `run.platform` instead of `run.scraper&.platform`.

## Testing done

Specs added (`spec/lib/morph/platform_file_spec.rb`, plus a new `#platform` block in `spec/models/run_spec.rb`) but **not executed** in this session - didn't want to stand up the full docker-compose/mysql/redis/elasticsearch stack for a change this size. Traced the logic by hand against the actual source and syntax-checked every touched file. CI should be the real gate here - flagging this clearly rather than claiming it's verified when it isn't.

Opening as a draft for the same reason.

## Summary by CodeRabbit

* **New Features**
  * Runs without an attached scraper can now determine their platform from the uploaded code.
  * Platform configuration is read consistently, including support for legacy `early_release` values.

* **Bug Fixes**
  * Improved platform selection when starting runs, including API-triggered runs without an associated scraper.

* **Tests**
  * Added coverage for platform-file handling, legacy platform mapping, and runs without scrapers.


benrfairless:
@coderabbitai full review

benrfairless (changes requested):
@Br3nda the bugs raised by coderabbit appear to be genuine. Can you review and re-request approval.

PR #1535, opened 2026-09-11
Update Rails to 7.0.8.7

benrfairless:
## What and why

Phase 2 Rails hop from doc/UPGRADE_PATH.md: Rails 6.1.7.10 to 7.0.8.7 (the final 7.0.x release), keeping the Sprockets asset pipeline by adding `sprockets-rails` explicitly now Rails no longer depends on it. Framework defaults stay at 6.1 here; #1536 adopts the 7.0 defaults one flag at a time.
ActiveAdmin moves 2.12.0 → 3.5.2 because 2.x does not support Rails 7.0, which also lifts the floor past the CVE-2023-51763/CVE-2024-37031 fixes the old Gemfile comment tracked separately; its Ransack 4 dependency now requires searchable-attribute allowlists, so `ApplicationRecord` restores the previous allow-everything behaviour (Ransack is only reachable through the admin-only ActiveAdmin UI). `rails app:update` output is adopted selectively, keeping the selective framework requires, SMTP settings, the #1498 deprecation-raise guard and the assets/inflections customisations, and dropping the generator's Active Storage artefacts (#1497 removed Active Storage).

- Part of #1355 (completed by #1536)
- Mined from the stale `chore/rails_70_and_ruby_30` branch (PR #1424) rather than rebasing it.
- Deprecation fixes: `to_s(:human_size)` → `ActiveSupport::NumberHelper` in the DB backup tasks; `redirect_back(fallback_location:)` → `redirect_back_or_to` via RuboCop's Rails/RedirectBackOrTo.
- schema.rb is regenerated in the 7.0 dump format (versioned header, explicit datetime precision): no schema change, only dump format.

## Type of change

- [x] Chore

## How this was checked

In the containerised dev environment: CI profile 854 examples / 0 failures (coverage 81.46% against the 80.05 threshold), `rubocop` and `haml-lint` clean, `srb tc` at 101 errors with the one new error in regenerated gem RBIs rather than app code (not CI-enforced), zero Rails deprecation warnings under the #1498 guard.
CodeRabbit review of the full #1534-#1537 stack diff: 0 findings.

- [x] Ran the automated tests
- [x] Ran `/code-review` (or equivalent): fixed any concerns (and rechecked), linked follow-up issues above, or noted why they are not important above
- [x] Documentation updated, or not needed (AGENTS.md version-pins section)

Assisted-by: OpenCode:anthropic.claude-fable-5

PR #1529, opened 2026-09-11
Adopt the Rails 6.1 framework defaults

benrfairless:
## What and why

Second half of the Phase 1 Rails hop (doc/UPGRADE_PATH.md): a fresh `rails app:update`, then the Rails 6.1 framework defaults adopted one flag per commit for bisectability, ending with `config.load_defaults 6.1` and the defaults file deleted.
The adoption commit keeps our customisations where the generator clobbered them (routes guard, selective framework requires, environment SMTP settings, the #1498 deprecation-raise guard, `:hybrid` cookie serializer) and drops the generator's Active Storage artefacts, since #1497 removed Active Storage and the engine is not loaded.
The one flag needing real thought is `cookies_same_site_protection = :lax`: the GitHub OAuth callback is a top-level redirect so Lax keeps sign-in working, but the runbook below verifies it explicitly after deploy.

- Part of #1354
- Stacked on #1528; this PR targets that branch and should be retargeted to `main` (or will auto-retarget) once #1528 merges.
- Not adopted: the cookie serializer stays `:hybrid`; the `:json` migration is Phase 2 work already done on the `chore/rails_70_and_ruby_30` branch.
- `config/boot.rb` keeps `require "logger"` with a corrected comment: the old TODO said to remove it on 6.1, but concurrent-ruby >= 1.3.5 breaks ActiveSupport below 7.1 without it (verified: boot fails when removed).

## Type of change

- [x] Chore

## How this was checked

Every per-flag commit ran the quick profile green (842 examples / 0 failures) in the containerised dev environment, so each commit is a deployable state.
Final tree: CI profile 854 examples / 0 failures (coverage 81.42% against the 80.05 threshold), `rubocop` and `haml-lint` clean, `srb tc` at 100 errors, unchanged from the base branch with none introduced.
CodeRabbit review of the full diff: 0 findings.

- [x] Ran the automated tests
- [x] Ran `/code-review` (or equivalent): fixed any concerns (and rechecked), linked follow-up issues above, or noted why they are not important above
- [x] Documentation updated, or not needed

## Deploy runbook (after merge, once #1528 has soaked 2-3 quiet days)

1. Deploy in a low-usage window: `cap production deploy`.
2. Sign out and sign back in via GitHub to verify OAuth under SameSite=Lax, in a fresh browser session.
3. Watch Sentry; `cap production deploy:rollback` is the fallback.
[...]

Issue #1358, opened 2025-09-30
Update morph to Rails 7.1.x in staging then production


ianheggie-oaf:
Update morph to Rails 7.1.x

Leave at ruby 3.0.7 so it should work on current production server as well as proposed new server.

Consider upgrading ruby version once we are on new server since [Ruby & Rails Compatibility Table](https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html) lists Rails 7.1 as requiring Ruby >= 2.7.0 and recommends 3.4.Z.

ianheggie-oaf:
@benrfairless - should we go down this track, or move to running scrapers on aws directly? there is enough effort needed to bring morph and buildstep up to scratch security wise, I wonder if its time to replace instead with a simpler system.

See:
* https://github.com/openaustralia/oaf-internal/issues/308

benrfairless:
We have decided that https://github.com/openaustralia/oaf-internal/issues/308 will not be going ahead.

I'm not for/against this at the moment - lets park this for discussion.

PR #1536, opened 2026-09-11
Adopt the Rails 7.0 framework defaults

### LINKED ISSUE #1355: Update morph to Rails 7.0.x

benrfairless:
## What and why

Adopts the Rails 7.0 framework defaults one flag per commit, ending with `config.load_defaults 7.0`, mirroring how #1529 adopted the 6.1 defaults. Most flags are inert here and each commit body says why; the ones needing real thought are below.

- Closes #1355 (with #1535)
- `key_generator_hash_digest_class = SHA256` ships with the cookie rotator from the [7.0 upgrade guide](https://guides.rubyonrails.org/v7.0/upgrading_ruby_on_rails.html): without it every session and remember-me cookie issued before the deploy becomes unreadable and everyone is logged out mid-soak. Remove the rotator after a soak, once live sessions have been rewritten.
- `cookies_serializer = :json` resolves the FIXME from the :hybrid migration, which has been rewriting Marshal cookies as JSON in production since November 2025.
- `cache_format_version` is the one deliberate hold-back, kept at 6.1 explicitly in `application.rb`: Rails 6.1 cannot read 7.0-format cache entries and production memcached outlives a `cap production deploy:rollback`, so the new format would break rollback during the soak. Flip it after the soak, or in the Rails 7.1 hop where the 7.0 format becomes the floor.
- `raise_on_open_redirects` needed one exemption: the Discourse SSO endpoint, whose whole job is redirecting to help.morph.io, gets `allow_other_host: true`. Everything else redirects internally (#1438 hardened the sign-out redirect separately).
- The Active Storage flags are dropped rather than enabled (#1497 removed Active Storage).

## Type of change

- [x] Chore

## How this was checked

In the containerised dev environment: CI profile 854 examples / 0 failures (coverage 81.46% against the 80.05 threshold), `rubocop` and `haml-lint` clean, `srb tc` unchanged at 101 (not CI-enforced), zero Rails deprecation warnings under the #1498 guard. The open-redirect flag initially broke the 22 Discourse SSO specs (tagged slow, so only the CI profile catches them), which is what prompted the `allow_other_host` exemption.
CodeRabbit review of the full #1534-#1537 stack diff: 0 findings.

- [x] Ran the automated tests
- [x] Ran `/code-review` (or equivalent): fixed any concerns (and rechecked), linked follow-up issues above, or noted why they are not important above
- [x] Documentation updated, or not needed

## Deploy runbook (after merge, as part of the Phase 2 deploy)

1. Deploy in a low-usage window: `cap production deploy`.
2. Verify GitHub sign-in explicitly (cookie serializer and key-digest change both touch session cookies).
3. Watch Sentry; `cap production deploy:rollback` is the fallback (rollback stays safe because cache_format_version is held at 6.1).
4. After the week-long Phase 2 soak: remove `config/initializers/cookie_rotator.rb` and flip `cache_format_version` to 7.0, or fold both into the Rails 7.1 hop (#1358).

Assisted-by: OpenCode:anthropic.claude-fable-5


ianheggie-oaf (approved):
Approved - makes sense to me

